PR #104, opened 2025-10-29
fix: generalize docker buildx upgrade

JGuinegagne:
This PR generalizes the fix for `docker buildx` install of #102 .

Also in this PR:
- improve inline help for `var.oauth_allowed_org` (leave blank instead of `""` which causes issues)
- improve inline help for `var.oauth_app_client_id` by providing direct links instead of `click here, click there` instructions
- fix typo in waiter script

### Testing
- deployed base template from scratch


michaelnchin (approved):
LGTM!

Issue #122, opened 2026-01-15
Persist Let's Encrypt Certificates Across Instance Replacements


JGuinegagne:
# Persist Let's Encrypt Certificates Across Instance Replacements

## Problem

E2E tests fail after GPU instance type switch because Let's Encrypt certificates in `acme.json` are lost during instance replacement, causing Traefik to hit rate limits (5 certs per 168h).

**Root Cause**: `/opt/docker/acme.json` is stored on the root EBS volume. When switching between CPU and GPU instances, **Terraform must replace the entire instance** because the **AMI must change**:

- CPU instances use Amazon Linux 2023 AMI
- GPU instances use Deep Learning AMI (pre-installed with NVIDIA drivers, CUDA, cuDNN, ML frameworks)

**AMI is immutable** - you cannot change an instance's AMI after launch. When Terraform detects the `ami_id` change, it must destroy the old instance and create a new one with the new AMI, which creates a new root EBS volume and loses all data including certificates.

The root volume size also changes (30 GB for CPU → 75 GB for GPU) as a consequence of the AMI change, but the **AMI change itself** is what forces the replacement.

**Note**: Instance type changes that use the same AMI (e.g., t3.medium → t3.xlarge, both using AL2023) preserve the instance and root EBS volume without triggering certificate loss.

```bash
$ jd host exec -- jq ".letsencrypt.Certificates | length" /opt/docker/acme.json
0  # Empty - certificates lost

$ jd server logs -s traefik
[ERR] Unable to obtain ACME certificate for domains
error="acme: error: 429 :: rateLimited :: too many certificates (5) already issued"

$ jd server status
Jupyter server status: FETCHING_CERTIFICATES  # Stuck forever
```

## Proposed Solution

Persist `acme.json` in **AWS Secrets Manager** and restore on instance startup.

### Implementation

**1. IAM Permissions** (extend instance role)
```hcl
secretsmanager:GetSecretValue
secretsmanager:UpdateSecret
# Resource: arn:aws:secretsmanager:*:*:secret:jupyter-deploy/${postfix}/acme-json*
```

**2. Startup Restore** (`docker-startup.sh.tftpl`)
```bash
# Try to restore from secret and sync to local file
if aws secretsmanager get-secret-value --secret-id "$SECRET_ID" 2>/dev/null; then
    aws secretsmanager get-secret-value --secret-id "$SECRET_ID" \
        --query 'SecretString' --output text > /opt/docker/acme.json
[...]

Issue #119, opened 2026-01-15
Externalize SSM Document Content to S3


JGuinegagne:
# GitHub Issue: Externalize SSM Document Content to S3

## Problem

The AWS SSM document `instance_startup` has a hard limit of 65,536 bytes. Currently, all startup scripts, docker files, and configuration files are embedded inline in the SSM document content, which causes the document to approach or exceed this limit.

### Current Size Analysis

With a baseline deployment (no external volumes):
- SSM document size: ~65KB (close to limit)

When adding multiple external volumes (2 EBS + 1 EFS):
- SSM document size: **69,020 bytes** (exceeds limit by 3,620 bytes)

### Impact

1. **E2E tests for external volumes are blocked** - Cannot test deployments with multiple external volumes
2. **Limited scalability** - Cannot add more features/scripts without hitting the limit
3. **Future feature additions will be constrained** - Any new functionality that requires additional scripts or config will push against this limit

## Root Cause

The SSM document (`aws_ssm_document.instance_startup` in `services.tf`) embeds all content inline:

1. **SaveUtilityScripts step** - Embeds management scripts:
   - `check-status-internal.sh`
   - `get-auth.sh`
   - `update-auth.sh`
   - `update-server.sh`
   - `refresh-oauth-cookie.sh`

2. **SaveDockerFiles step** - Embeds all service files:
   - Docker compose files
   - Dockerfiles (jupyter, oauth2-proxy, traefik, logrotator)
   - Jupyter startup scripts (uv and pixi variants)
   - Jupyter server config files
   - Fluent-bit logging config
   - All other service configuration

3. **CloudInit and volume mount scripts** - Additional bash scripts embedded inline

### Architecture: S3-Based Script Distribution

**Upload scripts/files to S3 during Terraform apply, download them on instance startup**

#### Implementation Details

1. **New S3 Bucket Module**

   Create new module `s3_bucket/` with the following files:

   **`s3_bucket/main.tf`**:
   ```hcl
   resource "aws_s3_bucket" "startup_scripts" {
     bucket_prefix = var.bucket_prefix
     force_destroy = var.force_destroy
   }

   resource "aws_s3_bucket_versioning" "startup_scripts" {
     bucket = aws_s3_bucket.startup_scripts.id
     versioning_configuration {
       status = "Enabled"
     }
   }

   resource "aws_s3_bucket_server_side_encryption_configuration" "startup_scripts" {
[...]

PR #124, opened 2026-01-16
feat: s3 config tls backup

JGuinegagne:
This PR 1/ implement s3-based scripts/service config files to avoid hitting the 65kB limit of SSM Document, 2/ backs up the TLS certs to an AWS secret, and pull them on instance restart.

Addresses #119 and #122 

Also in this PR:
- fix issues in terraform files logic where EBS volume would get replaced too often (because they used to depend on calculated AZ)
- fix permission issues on pre-existing volume for container on instance swap (we could not see it because the volumes were ALWAYS replaced!!)
- remove swap file from cloudinit
- update root volume size to a/ add more buffer (for docker runtime, logs, etc), b/ better scale w/ the AMI requirement

### User experience
- no change, it should be invisible to the users.

### Implementation
- add an S3 bucket (w/ encryption at rest w/ KMS, versioning, etc), grant permission to the EC2 IAM role
- add an AWS secret for the TLS certs, grant permission to the EC2 IAM role
- upload `cloudinit` and `volume_init` scripts to pull script/config files from S3
- add script to attempt to pull from secret at instance startup (to restore the certs)
- add script to upload the certs after instance startup (for the first time)
- add logic to retrieve the existing instance AZ, and attempt to preserve it. Otherwise, use the AZ of the 1st subnet of the default VPC as before
- increase timeouts for `docker-compose` (GPU images take a while to build because of `pytorch` dependencies), and of waiter script. Make timeouts depend on whether build is required.
- un-skip external volume E2E tests

### Testing
- tested manually upgrading & downgrading instances
- running E2E tests

### DIFF
--- a/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/notebook.py
+++ b/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/notebook.py
@@ -123,6 +123,29 @@ def run_notebook_in_jupyterlab(page: Page, notebook_path: str, timeout_ms: int =
     page.reload()
     page.wait_for_load_state("networkidle")
 
+    # Close all open notebook tabs to avoid strict mode violations when multiple notebooks are open
+    # This can happen if previous tests left tabs open due to errors
+    # Strategy: Use Command Palette to execute "application:close-all"
+    try:
+        # Close any open dialogs first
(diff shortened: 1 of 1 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

Issue #123, opened 2026-01-15
Add HTTP Readiness Check to Server Status


JGuinegagne:
# Add HTTP Readiness Check to Server Status

**Prerequisite**: https://github.com/jupyter-infra/jupyter-deploy/issues/119

## Problem

E2E tests get 404 errors even though server status reports `IN_SERVICE`. The status check only verifies containers are running and certs exist, but doesn't verify **Jupyter is actually responding to HTTP requests**. Containers can be "running" while Jupyter initializes inside.

## Proposed Solution

Add HTTP readiness check to `check-status-internal.sh` before reporting `IN_SERVICE`.

```bash
check_jupyter_ready() {
  # Try to reach the Jupyter API endpoint
  if docker exec jupyter curl -sf "http://localhost:8888/api" >/dev/null 2>&1; then
    return 0
  else
    log_message "Containers running but Jupyter not ready yet"
    return 1
  fi
}

# In main(), update the status logic:
elif [ "$container_status" -eq 0 ] && [ "$cert_status" -eq 0 ]; then
    if check_jupyter_ready; then
        log_message "In-service: containers running, certs available, Jupyter responding"
        exit 0  # IN_SERVICE
    else
        exit 40  # STARTING (containers up but Jupyter not ready)
    fi
else
    exit 50  # OUT_OF_SERVICE
```

**Update `get-status.sh`** to map new status codes:
```bash
case $STATUS in
  0)  echo "IN_SERVICE" ;;
  10) echo "INITIALIZING" ;;
  20) echo "STOPPED" ;;
  30) echo "FETCHING_CERTIFICATES" ;;
  40) echo "STARTING" ;;           # New: containers up, Jupyter initializing
  50) echo "OUT_OF_SERVICE" ;;     # Updated from 40
  *)  echo "UNKNOWN" ;;
esac
```

## Implementation

1. Add `check_jupyter_ready()` function to `check-status-internal.sh`
2. Update `main()` status logic to include Jupyter readiness check
3. Update exit codes in `check-status-internal.sh`: 40=STARTING, 50=OUT_OF_SERVICE
4. Update status code mapping in `get-status.sh` as shown above

## Files to Modify

- `services/commands/check-status-internal.sh` - Add `check_jupyter_ready()` and update exit codes
- `services/commands/get-status.sh` - Update status code mappings (40=STARTING, 50=OUT_OF_SERVICE)

Issue #121, opened 2026-01-15
Misleading "Authorization Failure" page shown for Jupyter application errors


JGuinegagne:
# Misleading "Authorization Failure" page shown for Jupyter application errors

## Summary

When Jupyter application fails (e.g., broken environment after `pixi remove jupyterlab`), users see an "Authorization Failure" error page even though they are properly authorized. This is misleading because the actual problem is a broken Jupyter service, not an authentication issue.

### Case 1: Broken Jupyter Environment (Misleading Error)

1. Start a working deployment
2. Break the Jupyter environment:
   ```bash
   # For pixi package manager
   jd server exec -- pixi remove --pypi jupyterlab

   # For uv package manager
   jd server exec -- uv remove jupyterlab
   ```
3. Refresh browser (without restarting container)
4. **Result:** "Authorization Failure" page is shown

**Actual cause:** Jupyter container is running but Jupyter application is broken and returns 500 errors.

### Case 2: Actual Authorization Failure (Correct Error)

1. Attempt to access with unauthorized GitHub account
2. **Result:** "Authorization Failure" page is shown

**Actual cause:** OAuth callback fails with "missing github user" error.

### Broken Jupyter Flow (Case 1)

**Jupyter logs (stderr):**
```
[E 2026-01-15 19:44:44.056 ServerApp] Uncaught exception GET /lab (172.18.0.3)
...
jinja2.exceptions.TemplateNotFound: 'index.html' not found in search path
...
[E 2026-01-15 19:44:44.073 LabApp] 500 GET /lab
```

**OAuth logs:**
```
15.248.7.139 ... jonathan.guinegagne@gmail.com [2026/01/15 19:44:28] ... 202 0 0.000
```
User IS authorized (202 response).

**Traefik logs:**
```
15.248.7.139 - - [15/Jan/2026:19:44:44 +0000] "GET /lab HTTP/2.0" 500 3460 "-"
"-" 337 "jupyter@docker" "http://172.18.0.5:4180" 22ms
```
500 error served from OAuth container (port 4180) via `oauth-faults` middleware.

### Actual Authorization Failure Flow (Case 2)

**OAuth logs (stderr):**
```
[2026/01/15 19:53:00]  Error creating session during OAuth2 callback: missing github user
```

**OAuth logs (stdout):**
```
15.248.7.139 ... "/oauth2/callback?..." HTTP/1.1 500 2772 1.268
[...]

Issue #93, opened 2025-09-16
Harden the server status script


JGuinegagne:
`jd server status` calls [this script](https://github.com/jupyter-ai-contrib/jupyter-deploy/blob/5a8d9886472e9ef8adc6ac4950631187a688c081/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/services/commands/check-status-internal.sh#L30) which checks if docker container services are running.

However, in some cases a container can build & start, but then fail on its startup script and get into a loop of restart. The issue is that the startup tracker script, which calls the status command [here](https://github.com/jupyter-ai-contrib/jupyter-deploy/blob/5a8d9886472e9ef8adc6ac4950631187a688c081/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/engine/local-await-server.sh.tftpl#L186) can get tripped by this and believe the services are up. When that happens, if a user does `jd open`, they get a `404` because `jupyter` service is unresponsive. Example where this caused an issue: #92 

We should ensure that the `jupyter` container is indeed up and has finished running before reporting `InService`. As P0, waiting for the service to have been up for 1 minute might be sufficient.

PR #125, opened 2026-01-20
fix: envs consistency, waiter & hash logic

JGuinegagne:
This PR addresses more issues with the base template that I discovered with the e2e tests:
1. add docker health check to `jupyter` container, hitting `localhost:8888/lab/favicon.ico` to avoid the expensive `/lab`. Note that `/api` didn't capture lab failure
2. add traefik health check on `jupyter` service, which redirects to a new `503` page if the request is authorized but the jupyter service isn't healthy
3. add hash capturing all scripts & docker configuration files to ensure `jd config && jd up` picks up and upload any local changes to these files to the instance
4. change image hash logic to capture other images and re-trigger their build as needed (log rotator in this case)
5. restart services instead of naively starting in the SSM document
6. remove `torch` and `numpy` dependencies from uv and pixi env for gpu, remove `ipywidgets` from pixi env - this greatly accelerates image builds, and users can install `torch` themselves as needed.
7. fix regression introduced by #124

Addresses #93, #121, #123

Also in this PR:
- add "configuration test" to let LLMs verify that their changes to any `terraform` files are valid; this runs `jd init /tmp/some/subdir/` then `jd config` on it and assert that it works - update `CLAUDE.md` to explain to the LLM how to run it. Then users can just instruct "run the configuration tests" and the LLM knows what to do
- update `CLAUDE.md` to provide specific commands and information about how to debug issues.

### User experience
- if the `jupyter` service becomes unavailable for any reason (trigger this with `uv remove jupyterlab` and browser refresh), an authorized user now sees a dedicated html page
- when a user updates any script or docker configuration file in an already deployed project, they can just run `jd config && jd up` to apply them

### Implementation
- add health checks to `docker-compose.yml` (both native docker and traefik)
- add new html template for 503 page, update redirect logic
- add `deployment_scripts_hash` output, trigger recreate on SSM-instance association if it changes
- update `jupyter_build_hash` > `images_build_hash` and capture log rotator dockerfile and script
- use `sh update-server.sh restart all` instead of `sh docker-startup.sh` at the end of startup SSM document
- remove previous instance tracking in `main.tf` (failed on greenfield projects in some cases)
- remove `torch`, `numpy` and conditional on hardware type in `pyproject.toml` and `pixi.toml`
- update plugin to support configuration tests

### Testing
- run e2e tests against existing deployment: `just test-e2e sandbox "" mutate=true`
- run brand new deployment: `just test-e2e sandbox-e2e "" mutate=true`

### DIFF
--- a/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_gpu.py
+++ b/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_gpu.py
@@ -85,6 +85,7 @@ def test_run_gpu_notebook(
 
     # Restart server to ensure clean session (prevents "Document session error" dialogs)
     e2e_deployment.cli.run_command(["jupyter-deploy", "server", "restart"])
+    e2e_deployment.ensure_server_running()
 
     # Re-authenticate after server restart
     github_oauth_app.ensure_authenticated()
--- a/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_pixi.py
+++ b/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_pixi.py
@@ -1,6 +1,5 @@
 """E2E tests for Pixi package manager functionality."""
 
-import time
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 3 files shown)


michaelnchin (dismissed):
LGTM, thanks @JGuinegagne !

PR #136, opened 2026-01-26
fix: surface underlying cmd retcode in exec cmds

### LINKED ISSUE #129: Fix jd server exec and jd host exec to propagate command exit codes
**Problem:** `jd server exec` and `jd host exec` always exit with code 0 even when the target command fails. This caused EFS test failures to be detected late during file verification instead of immediately at the upload command.

**Root Cause:** The manifest commands don't capture `ResponseCode` from SSM command execution, only stdout/stderr.

**Fix:** Update 3 components for both `server.exec` and `host.exec`:

1. **Manifest** (`libs/jupyter-deploy-tf-aws-ec2-base/.../manifest.yaml`): Add `ResponseCode` to results
2. **Handler** (`libs/jupyter-deploy/jupyter_deploy/handlers/resource/*_handler.py`): Return exit code as 3rd tuple element
3. **CLI** (`libs/jupyter-deploy/jupyter_deploy/cli/*_app.py`): Propagate exit code with `raise typer.Exit(code=exit_code)`

**Note:** This may break negative E2E tests like `test_server_exec_failed_command` which expect specific error handling behavior and will need updating.

JGuinegagne:
This PR update `jd host exec` and `jd server exec` logic to look through the command executed on the host or server and surface its return code. Previously, the exec commands returned `0` as return code if the `SSM:SendCommand` API call succeeded (for the base template), irrespective of whether the underlying command failed or succeeded. This behavior was counter-intuitive for users.

Addresses #129 

### User experience
- `jd server exec -- whoami` --> `jovyan`, retcode=0
- `jd server exec -- unknown-cmd` --> <error msg>, retcode=126 (docker)
- `jd host exec -- pwd` --> `/`, retcode = 0
- `jd host exec -- unknown-cmd` --> <error msg>, retcode=127

### Implementation
- update `HostHandler.exec()` and `ServerHandler.exec()` to return the retcode
- update `host`/`server` cli to return the underlying retcode from handler
- update the base template `manifest.yaml` to add the retcode to the `exec` cmds results
- update e2e tests and plugin to take into account this new behavior

### Testing
- unit tests (updated)
- ran e2e tests

### DIFF
--- a/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/manifest.yaml
+++ b/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/manifest.yaml
@@ -93,6 +93,9 @@ commands:
   - result-name: host.exec.stderr
     source: result
     source-key: '[0].StandardErrorContent'
+  - result-name: host.exec.returncode
+    source: result
+    source-key: '[0].ResponseCode'
 - cmd: server.status
   sequence:
   - api-name: aws.ssm.wait-command-sync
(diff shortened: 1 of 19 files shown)


michaelnchin (approved):
Thanks @JGuinegagne !

PR #139, opened 2026-02-02
fix: harden ebs mount, instance restart

### LINKED ISSUE #138: Add retry logic and Terraform dependencies for EBS volume attachment during boot
## Problem

E2E tests `test_external_volumes_provisioning` and `test_upgrade_config` fail intermittently because external EBS volumes are not mounted inside the Jupyter container. The volumes are attached to the instance but the mount points don't exist in `/home/jovyan/`.

## Root Cause

Race condition between EBS volume attachment and SSM startup script execution:

1. **Missing Terraform Dependency** - The SSM association in `services.tf:519-522` doesn't wait for EBS volume attachments before running the startup script
2. **No Device Availability Retry** - In `cloudinit-volumes.sh.tftpl:48-50`, when the device isn't found, the script just logs an error and continues instead of retrying or failing
3. **Silent Failure** - Without proper failure handling, Docker services start with missing volume mounts

### Missing Terraform Dependencies
**File:** `libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/engine/services.tf:519-522`

The SSM association `depends_on` block only includes:
```terraform
depends_on = [
  module.secret,
  module.ec2_instance,
]
```

It should also wait for volume attachments to complete.

### No Retry Logic and Silent Failure for EBS Devices
**File:** `libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/services/cloudinit-volumes.sh.tftpl:48-50`

When device not found:
```bash
else
    echo "Error: Device $device_name not found for mounting to $FULL_MOUNT_POINT"
fi
```

This has two problems:
1. **No retry** - Doesn't wait for device to appear (unlike EFS which has retry logic at lines 59-91)
2. **Silent failure** - Just logs an error without calling `exit 1`, so script continues and Docker starts without the mount

### Fix 1: Add Terraform Dependencies
Add volume attachments to the SSM association's `depends_on` block:
```terraform
depends_on = [
  module.secret,
  module.ec2_instance,
  module.volumes.aws_volume_attachment.jupyter_data_attachment,
  module.volumes
]
```

### Fix 2: Add EBS Device Availability Retry AND Explicit Failure
This fix addresses both problems:
1. **Add retry logic** - Wait for device with exponential backoff (similar to EFS utils installation)
2. **Exit on failure** - Call `exit 1` if device not found after retries (instead of silently continuing)

```bash
# Wait for device to become available with retry
MAX_RETRIES=5
RETRY_COUNT=0
RETRY_DELAY=2

while [ $RETRY_COUNT -le $MAX_RETRIES ]; do
    if [ -b "$device_name" ]; then
        echo "Device $device_name is available"
        break
    else
        RETRY_COUNT=$((RETRY_COUNT + 1))
        if [ $RETRY_COUNT -le $MAX_RETRIES ]; then
            echo "Device $device_name not yet available (attempt $RETRY_COUNT/$((MAX_RETRIES + 1))), waiting ${RETRY_DELAY}s..."
            sleep $RETRY_DELAY
            RETRY_DELAY=$((RETRY_DELAY * 2))
        else
            echo "Error: Device $device_name not found after $((MAX_RETRIES + 1)) attempts"
            exit 1  # Explicit failure instead of silent continue
        fi
    fi
done
```

## Expected Behavior After Fixes

1. Terraform dependency ensures EBS volumes are attached before SSM runs
2. If there's still a timing issue (device node not in `/dev/`), the retry handles it
3. If device is truly unavailable after retries, script **exits with error code** (not silent)
4. With `onFailure: exit` in SSM document (from #128), startup fails explicitly
5. Users see clear failure rather than silent missing mounts

## Related Issues

- #130 - EFS utils installation transient failures (fixed with retry logic)
- #128 - SSM document cascade failures (fixed with `onFailure: exit`)

This issue addresses the same type of race condition for EBS volumes.

JGuinegagne:
This PR a/ adds retry logic in volume initialization script responsible to configure external mount on the host, b/ make the ssm-association dependent on the completion of ebs creation and mount (as in, the aws operation), c/ add handling for stopped instances in waiter script.

Addresses #138 

### Customer experience
- no change

### Implementation
- add retries to `volume-init`
- add `module.volumes` as dependency of SSM Association
- update waiter script

### Testing
- run E2E test
- run `jd config & jd up` from stopped instance --> verified restart

### DIFF
--- a/env.example
+++ b/env.example
@@ -3,7 +3,16 @@
 # The docker-compose will automatically load .env from the repository root
 
 # ==============================================================================
-# DEPLOYMENT CONFIGURATION
+# TEST CONTAINER BUILD CONFIGURATION
+# ==============================================================================
+
+# Leave blank the host settings, the justfile methods will populate them
+HOST_UID=
+HOST_GID=
+
+
+# ==============================================================================
+# JUPYTER DEPLOYMENT CONFIGURATION
 # ==============================================================================
 
 # Domain configuration
--- a/justfile
+++ b/justfile
@@ -25,22 +25,31 @@ e2e-compose-file := "libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/image/docker-
 e2e-image-name := "jupyter-deploy-e2e-aws-ec2-base"
 e2e-image-tag := "latest"
 
-# Build E2E test image (optional - docker compose will build automatically on first up)
-e2e-build:
-    @echo "Building E2E test image with {{container-tool}}..."
-    {{container-tool}} compose --project-directory {{justfile_directory()}} -f {{e2e-compose-file}} build --no-cache
+# Start E2E container in background (always builds to ensure correct UID/GID)
+# Usage: just e2e-up [no-cache=true]
+e2e-up no_cache="false":
+    #!/usr/bin/env bash
+    set -euo pipefail
 
-# Start E2E container in background (builds image if needed)
-e2e-up:
-    @echo "Starting E2E container (will build image if needed)..."
-    @mkdir -p {{justfile_directory()}}/test-results
-    @chmod 777 {{justfile_directory()}}/test-results
-    @echo "Setting AWS credentials permissions for container access..."
-    @chmod 644 ~/.aws/credentials ~/.aws/config 2>/dev/null || true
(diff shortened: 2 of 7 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #153, opened 2026-02-09
fix: decouple terminal display from core logic for error handling

JGuinegagne:
This PR decouples error display from the core logic, and moves display handling to `cli/` module.

This is the first phase to decouple core handlers from the specific interface (CLI). The goal is to ensure the architecture is modular, so that we can easily separate the CLI from the core handling in the future. Ultimately, we should be able to support several interfaces, for example to serve a UI in addition to a CLI.

The first milestone consists in ensuring that the core of the package (handlers, engine, providers, core utils methods) do not depend on CLI and terminal-specific display dependencies: `rich` & `typer` mainly. The second milestone will be to remove `console.print()` statement from handlers for outcomes (e.g. in `OpenHandler` [method](https://github.com/jupyter-infra/jupyter-deploy/blob/7e1692ee49487ccdb579de97d8b2e4c8674aa5d1/libs/jupyter-deploy/jupyter_deploy/handlers/project/open_handler.py#L60)). The third milestone will consist in removing console dependencies in [InstructionRunner](https://github.com/jupyter-infra/jupyter-deploy/blob/7e1692ee49487ccdb579de97d8b2e4c8674aa5d1/libs/jupyter-deploy/jupyter_deploy/provider/manifest_command_runner.py#L43) implementations. That will require a new protocol to send incremental updates.

### Customer experience
- no change

### Implementation
- define a range of new custom errors in `<jd>/exceptions`
- add a `handle_cli_errors` in `<jd>/cli` module to catch and print them appropriately
- update the handler, utility and provider method to raise such exception instead of printing

### Testing
- manual testing against existing deployment
- run E2E tests

### DIFF
--- a/libs/jupyter-deploy/jupyter_deploy/cli/error_decorator.py
+++ b/libs/jupyter-deploy/jupyter_deploy/cli/error_decorator.py
@@ -9,6 +9,7 @@ from rich.console import Console
 from jupyter_deploy.exceptions import (
     ConfigurationError,
     DownAutoApproveRequiredError,
+    HostCommandInstructionError,
     IncompatibleHostStateError,
     InstructionNotFoundError,
     InteractiveSessionError,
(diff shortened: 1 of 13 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

michaelnchin (approved):

Issue #166, opened 2026-03-04
Add `s3-only` store type (S3 without DynamoDB locking)


JGuinegagne:
# Remove DynamoDB state locking from terraform backend

## Problem

The `s3-ddb` store type uses a DynamoDB table for terraform state locking (`dynamodb_table` in `backend.tf`). For a single-user deployment tool, this causes more problems than it solves:

- **Stale locks from interrupted commands**: if `jd up` is interrupted (Ctrl+C, network drop, terminal closed), terraform leaves a lock entry in DynamoDB. The next command fails with "state locked".
- **Hard to recover**: requires `terraform force-unlock <LOCK-ID>` from the engine directory.
- **Rare concurrent access in practice**: Audience is primarily a single-user managed project.

Auto force-unlock was considered but rejected: it cannot distinguish stale locks from legitimately-held locks, adding complexity without real safety.

## Solution: add `s3-only` store type

Add a new store type that uses S3 without DynamoDB locking:

- New `s3-only` store type: S3 bucket only, no DynamoDB table, no `dynamodb_table` in `backend.tf`
- Keep existing `s3-ddb` store type intact for users who want locking
- Switch the base template manifest from `s3-ddb` to `s3-only`
- `_BACKEND_TEMPLATE` in `tf_store_access.py` should be parameterized (with/without `dynamodb_table`)

The DynamoDB table code (`ensure_store`, `create_lock_table`, etc.) remains available in `s3-ddb` for users who want to leverage lock.

PR #167, opened 2026-03-04
feat: s3-only store

JGuinegagne:
This PR adds a second `store-type: s3-only`, which does not use the dynamoDB-based lock mechanism to prevent concurrent changes in the terraform state. It switches the base template to use this store mechanism, effectively storing the terraform state and project details only on the account S3 bucket.

This is necessary because the lock caused usability friction with no real benefit for a typical jupyter-deploy user. The more advanced case remain possible by specifying `store-type: s3-ddb`.

Closes: #166 

### User experience
- no change, this is invisible to the user, `jupyter-deploy` manages the remote-state on its own

### Implementation
- add an `S3StoreManager` class mapping to `store-type: s3-only`
- change `S3DynamoDbTableStoreManager` class to extend `S3StoreManager`
- add handling for both `StoreType` in `StoreManagerFactory`
- add a `StoreType` enum and a corresponding `InvalidStoreTypeError`, w/ handling in the cli decorator
- add handling for `store-type: s3-only` in `TerraformStoreAccessManager` to generate the appropriate `backend.tf`

### Testing
- tested manually against existing deployment
- delete the DDB table in the account, switched instance with `jd config --instance-type t3.xlarge` & `jd up` again

### DIFF
--- a/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/manifest.yaml
+++ b/libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/manifest.yaml
@@ -21,7 +21,7 @@ values:
   source: output
   source-key: persisting_resources
 project-store:
-  store-type: s3-ddb
+  store-type: s3-only
 services:
 - jupyter
 - traefik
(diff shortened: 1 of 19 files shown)


michaelnchin (commented):
Thanks @JGuinegagne - LGTM!

michaelnchin (approved):

PR #171, opened 2026-03-09
feat: mask secrets

JGuinegagne:
This PR implements the automatic masking of secrets that users may input in `variables.yaml`. Any local files that tracks secret value is gitignored, and will NOT be tracked in the remote store.

Also in this PR:
- remove stale `-s` flag in `jd config`, which wasn't applied anyway when secrets were specified in `variables.yaml`

### User experience
- users can still input their secret value in `variables.yaml`, then run `jd config`
- the value of the secret in `variables.yaml` will get changed to `****`
- any subsequent calls to `jd config` and `jd up` will use the stored secret
- the user can change the value in `variables.yaml`, and run `jd config` again

### Implementation
- the secret value that the user input gets saved in `<project-dir>/engine/jdinputs.secrets.auto.tfvars`
- terraform plan uses this file automatically
- add a `.mask_secrets()` method to `ConfigHandler`
-  this method parses `variables.yaml`, and replace any secret value with `****`
- call from `cli/app:config()` after `record()`

### Testing
- ran configuration tests
- tested manually (`jd config`)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #186, opened 2026-04-01
fix: e2e workflows

JGuinegagne:
This PR fixes permission and configuration issues on the GitHub workflows for e2e tests:
- bypass fetch secret for bot recovery code in `ci-restore` (the CI job does NOT have permission to read that secret)
- add the `permission` block in workflow files to allow `id-token: write`
- update ci template secret policy to no longer deny innocuous read/list operation (but still deny `GetSecretValue`), which is necessary for `terraform plan` (called in `ci-restore` via `jd config`)

Also in this PR:
- fixed `playwright` misconfiguration that led to running all browser-based E2E tests twice!!

### Testing
- pushed to a tmp branch in the main repo (`fix/e2e-workflows`)
- added a `.github/workflows/test-permission.yml` and pushed to trigger the e2e workflow:
```yaml
name: Test permissions (temporary)

on:
  push:
    branches:
      - fix/e2e-workflows

permissions:
  id-token: write
  contents: read

jobs:
  test-reusable:
    name: "Test reusable workflow permissions"
    uses: ./.github/workflows/e2e-base-job.yml
    secrets: inherit
    with:
      oauth-app-num: "1"
      test-filter: "test_application"
      timeout-minutes: 30
```
- verified that the test passed, see [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/23851543464/job/69532761173)

### DIFF
--- a/.github/workflows/pr-e2e-base-fresh.yml
+++ b/.github/workflows/pr-e2e-base-fresh.yml
@@ -25,6 +25,10 @@ on:
         type: string
         default: ""
 
+permissions:
+  id-token: write
+  contents: read
+
 concurrency:
   group: e2e-${{ inputs.oauth-app-num }}
   cancel-in-progress: false
--- a/.github/workflows/pr-e2e-base.yml
+++ b/.github/workflows/pr-e2e-base.yml
@@ -20,6 +20,10 @@ on:
         type: string
         default: ""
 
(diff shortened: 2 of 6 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #187, opened 2026-04-02
fix: E2E test sequence, webbrowser access in ci

JGuinegagne:
This PR fixes a few issues w/ E2E tests when running in CI:
- add a method to ensure `jd up` was run locally before assertions again `jd history` on up logs
- make webbrowser accessible in test container of the CI
- fix paper cut where `.jd/store.yaml` was not persisted on time when `jd up` was run once
- pin `terraform` version in e2e test container to ensure compatibility between runs and environments

### Testing
- pushed to other branch, verify [workflow](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/23881866335) w/ full non-mutating e2e tests
- ran non-mutating e2e tests locally

### DIFF
--- a/.github/workflows/e2e-base-job.yml
+++ b/.github/workflows/e2e-base-job.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@v3
         with:
-          terraform_version: "1.5.5"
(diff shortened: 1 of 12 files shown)


michaelnchin (approved):
Looks good - thanks @JGuinegagne !

PR #192, opened 2026-04-06
fix: secrets in oauth container, ci papercuts

JGuinegagne:
This PR fixes various minor issues in the CI runs:
1. add `.github/README.d` to explain how to setup the ci, including grant permission to the oauth apps
2. add `just` method to create a base project from CI, and to create a ci project
3. fix issues in terraform logic to:
    3.1. ensure the `cloudinit` SSM document updates whenever the auth (usernames, org, teams) are changed
    3.2. do NOT overwrite the auth settings on host restart (ie no-op in `cloudinit` if /etcd auth file already exists
4. add waiter for server to be actually ready in plugin util, use in `test_application`
5. fix env vars handling in `just setup-env-base` to ignore current app state and use the CI settings instead

Also in this PR:
1. tighten up `oauth` container security by passing secrets as file instead of env var (visible w/ `docker inspect oauth`)
2. pass `cookie_refresh=1h` setting which re-verify that the user still belongs to the approved orgs/teams

### Testing
- clean webbrowser context (`.auth`), run the org/teams, users and application test against canary app

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,0 +1,151 @@
+# CI Infrastructure Setup
+
+This document describes the one-time setup required for the E2E CI workflows.
+
+## Prerequisites
+
+- AWS account with the CI infrastructure template deployed (`jupyter-infra-tf-aws-iam-ci`)
+- A dedicated GitHub bot account for automated testing
+- GitHub OAuth apps (1-5) registered under the bot account
+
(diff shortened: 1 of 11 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #198, opened 2026-04-13
chore: upgrade deps

JGuinegagne:
This PR A/ upgrades the dependencies of all packages in the monorepo (`uv sync --all-packages --upgrade`), B/ unpins dependency in the CI to ensure that the CI tests against the setup a new user would get when running `pip install jupyter-deploy` (or `uv add jupyter-deploy`).

### Testing
- see CI [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/24348529703) against now-deleted branch

PR #207, opened 2026-04-23
fix: variables description

### LINKED ISSUE #206: Regression: variable descriptions show <<-EOT instead of actual text
# Regression: variable descriptions show `<<-EOT` instead of actual text

## Symptom

`jd show -v <name> --description` and `jd config` display `<<-EOT` for every
variable that uses a heredoc description. Outputs are unaffected because
`outputs.tf` uses inline strings (0 heredocs vs 28 in `variables.tf`).

## Root cause

`python-hcl2` v8 (upgraded from v7 in `f959c0c`) no longer expands heredoc
blocks. Instead it returns the raw heredoc markers as a quoted string literal:

```
v7:  "The EC2 instance type to use.\n..."
v8:  "\"<<-EOT\n    The EC2 instance type to use.\n  EOT\""
```

The `strip_hcl2_quotes()` compatibility shim (`tf_varfiles.py:11-36`) was
written to handle v8's other breaking change — outer quotes on plain strings
and literal-to-string conversions — but it does not detect or expand heredocs.

After quote-stripping the value becomes `<<-EOT\n    The EC2...\n  EOT`.
`get_trimmed_header()` then takes the first line, yielding `<<-EOT`.

## Affected code path

1. `hcl2.loads()` → raw heredoc string with markers
2. `strip_hcl2_quotes()` → strips outer `"`, leaves `<<-EOT...EOT` intact
3. `TerraformVariableDefinition.description` → stores the raw heredoc string
4. `get_cli_description()` → `get_trimmed_header()` → first line = `<<-EOT`

## Scope

- `jd show -v <var> --description` — shows `<<-EOT`
- `jd config` table — description column shows `<<-EOT`
- All 28 heredoc variables in the base template are affected
- Outputs and inline-string variables are not affected

JGuinegagne:
This PR fixes a regression introduced by #198 where `hcl2` upgrade to `9.*` broke the parsing of `variables.tf` description, so that `jd config --help` no longer showed a usable variable description. Also affected `jd show -v VARNAME --description`.

### Implementation
- fix `<jd>/engine/terraform/tf_varfiles` to correctly parse the content
- expand unit tests to catch future regressions (both for `variables.tf` and `outputs.tf`)
- add e2e test coverage for variables and outputs description in `jd config --help` and `jd show`

### DIFF
--- a/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_config_set.py
+++ b/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_config_set.py
@@ -31,6 +31,10 @@ def test_config_help_show_variables(e2e_deployment: EndToEndDeployment) -> None:
     for var in expected_variables:
         assert var in result.stdout, f"Expected variable '{var}' in config help output"
 
+    # Verify descriptions are expanded (not raw heredoc markers)
+    assert "<<-EOT" not in result.stdout, "Config help contains unexpanded heredoc marker <<-EOT"
+    assert "<<EOT" not in result.stdout, "Config help contains unexpanded heredoc marker <<EOT"
+
 
 @pytest.mark.cli
 def test_config_set_string_variable(e2e_deployment: EndToEndDeployment) -> None:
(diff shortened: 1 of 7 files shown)


bhanderig (approved):
LGTM

PR #213, opened 2026-04-29
feat: add username to ci template

JGuinegagne:
This PR adds a `username` variable to the CI template. That's needed for testing on personal GitHub account where the email prefix may diverge from the GitHub username. Also update the `just` method `just auth-bot-username` to read that ssm-parameter instead of inferring from the email

### Testing
- updated the CI template in my personal account
- verified the `just` method manually

### DIFF
--- a/scripts/config_base_from_ci.py
+++ b/scripts/config_base_from_ci.py
@@ -96,8 +96,13 @@ def main() -> None:
     if allowed_usernames_arg:
         allowed_usernames = allowed_usernames_arg
     else:
-        # Default: allow the bot account (extract username from email)
-        bot_username = bot_email.split("@")[0] if "@" in bot_email else bot_email
+        result = subprocess.run(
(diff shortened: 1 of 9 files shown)


scripts/config_base_from_ci.py
```diff
@@ -96,8 +96,13 @@ def main() -> None:
     if allowed_usernames_arg:
         allowed_usernames = allowed_usernames_arg
     else:
-        # Default: allow the bot account (extract username from email)
-        bot_username = bot_email.split("@")[0] if "@" in bot_email else bot_email
+        result = subprocess.run(
```
bhanderig, line 99: nit: A small helper would reduce duplication of this pattern. I see this being used the same way in few places

bhanderig (approved):

scripts/config_base_from_ci.py
JGuinegagne: Makes sense, lemme create a GH issue to track this

PR #223, opened 2026-05-18
fix: gh workflows papercuts

### LINKED ISSUE #190: CI: Transient setup-uv failure causes cascading job failure
# CI: Transient `setup-uv` failure causes cascading job failure

## Observed

Run [#23981223413 attempt 1](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/23981223413/job/69949031371) —
"Mutating: GPU / Run E2E tests" failed at step 3 ("Install UV and Python").

`astral-sh/setup-uv@v5` exited with code 1 during "Activating python venv..." after installing uv 0.7.8 successfully.
Because step 3 failed, step 4 (`just`) was never installed, causing a secondary `exit code 127: just: command not found`.

Succeeded on attempt 2 with no code changes — classic transient/race condition.

## Root cause (likely)

Transient failure in `setup-uv`'s venv activation — possibly a cache restore race or GitHub-hosted runner flake.

## Recommendation

Add a retry wrapper around the `install-uv` composite action (or inline the step with retries).
The simplest option: use [`nick-fields/retry@v3`](https://github.com/nick-fields/retry) around the uv install,
or add `continue-on-error: true` + a manual retry step in the composite action.

### LINKED ISSUE #220: CI: `just: command not found` in always() cleanup steps
When `extractions/setup-just@v2` has a transient failure, the `Export auth state` step
(`if: always()`) fails with `just: command not found`.

**Fix:** Add `id: install-just` to the install step and guard cleanup with:
`if: always() && steps.install-just.outcome == 'success'`

**Affected:** `e2e-base-job.yml`, `e2e-base-fresh.yml` (all `auth-export` steps)

**Hardening:** Replace `extractions/setup-just@v2` with a retryable shell install:
```yaml
run: curl --retry 3 --proto '=https' -fsSL https://just.systems/install.sh | bash -s -- --to /usr/local/bin
```

**Ref:** https://github.com/jupyter-infra/jupyter-deploy/actions/runs/25887248197/job/76090374704

JGuinegagne:
This PR addresses infrequent issues w/ GitHub workflows, and fixes README of `jupyter-deploy` to improve PyPI landing page display.

Closes: #190 
Closes: #220

### DIFF
--- a/.github/actions/install-just/action.yml
+++ b/.github/actions/install-just/action.yml
@@ -12,5 +12,8 @@ runs:
   steps:
     - name: Install just
       shell: bash
-      # yamllint disable-line rule:line-length
-      run: curl --retry 3 --proto '=https' -fsSL https://just.systems/install.sh | bash -s -- --tag ${{ inputs.version }} --to /usr/local/bin
+      # yamllint disable rule:line-length
+      run: |
+        curl --retry 3 --proto '=https' -fsSL https://just.systems/install.sh | bash -s -- --tag ${{ inputs.version }} --to /usr/local/bin
+        just --version
+      # yamllint enable rule:line-length
--- a/.github/actions/install-just/action.yml
+++ b/.github/actions/install-just/action.yml
@@ -1,0 +1,16 @@
+name: install-just
+description: 'Install just command runner with retry'
+
+inputs:
+  version:
+    description: 'just version to install'
+    required: false
+    default: '1.40.0'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install just
+      shell: bash
+      # yamllint disable-line rule:line-length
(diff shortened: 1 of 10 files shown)


.github/actions/install-just/action.yml
```diff
@@ -0,0 +1,16 @@
...
+inputs:
+  version:
+    description: 'just version to install'
+    required: false
+    default: '1.40.0'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install just
```
bhanderig, line 13: nit: since we're fixing flakiness here, should we also verify just actually installed? curl | bash can exit 0 on a truncated download or a failed write to /usr/local/bin, and you'd only find out a few steps later when just <recipe> errors with command not found. A quick just --version at the end of the action would catch it at the source.

bhanderig (approved):
LGTM

PR #244, opened 2026-06-09
fix: delete ws before down to avoid race conditions

JGuinegagne:
This PR adds a destroy null-resource step to delete all workspace resources that may reference other resources and therefore prevent their deletion: workspaces (may reference template or access strategies) and templates (may reference access strategies).

This prevents a deadlock destroy case observed in [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/27175437366/job/80223200948), where the `workspace_defaults` chart gets locked on template/access strategy deletion finalizers.

Also ignore failure to restore secret on delete action as observed in cascading failure [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/27175851111/job/80224445117).

### Implementation
- add a `destroy_workspaces` null resource in `<template>/engine/workspace.tf`
- tolerate failure on restore secret for delete operations in the CI

### Testing
- update a deployment in place (app 4), created a workspace manually, ran `jd down`


bhanderig (approved):
LGTM

PR #259, opened 2026-06-12
chore: harden ci, improve web-app e2e tests

### LINKED ISSUE #246: Move web app E2E fixtures to pytest-jupyter-deploy plugin
## Context

From [PR #230 review comment](https://github.com/jupyter-infra/jupyter-deploy/pull/230#discussion_r3358857498) by @JGuinegagne:

> Because the `jupyter-k8s-ui` is the standard entry point for multi-apps deployment, we should move these fixtures or utility to the plugin in `libs/pytest-jupyter-deploy`. That way, several templates can re-use the same navigation flow in their E2E tests.

## What

Move the web app navigation fixtures and utilities currently in `libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_web_app.py` into `libs/pytest-jupyter-deploy` so they can be shared across multiple template E2E test suites.

## Why

The UI is becoming the standard entry point for multi-app deployments. Other templates (beyond `aws-eks-oidc`) will need the same browser navigation flow for their E2E tests.

### LINKED ISSUE #251: E2E release gate (CLI): retry image build to absorb Test PyPI propagation race
## Problem

The `E2E release gate (CLI)` build jobs (`Build image (bare|aws|aws-k8s)`) intermittently fail when resolving the freshly-published package from Test PyPI.

Observed in [run 27367238721, job 80870958567](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/27367238721/job/80870958567): the `aws-k8s` build failed at the in-Dockerfile `uv sync` with:

```
× No solution found when resolving dependencies for split (markers: python_full_version >= '3.15'):
╰─▶ Because there is no version of jupyter-deploy[aws]==0.6.2 and your
    project depends on jupyter-deploy[aws]==0.6.2, ... unsatisfiable.
```

A re-run passed with no code change.

## Root cause

Test PyPI CDN eventual-consistency race — **not** a real defect:

- `0.6.2` was uploaded at 18:06:50; all three image builds started at 18:08:20 (~90s later).
- `bare` and `aws` (which also pull `jupyter-deploy` from Test PyPI) succeeded; only `aws-k8s` failed — it happened to query a stale `/simple/` CDN node that hadn't picked up the new file.
- Package metadata is correct (`requires-python >=3.12`, `aws`/`k8s` extras present).
- The `uv` "no solution for split python_full_version >= '3.15'" wording is a red herring — the universal resolver's way of saying "version not found on the index I queried", not a Python-version problem.

`release-cli.yml` already polls the Test PyPI **JSON** endpoint (`/pypi/<pkg>/<ver>/json`) before the gate, but the Docker build resolves against the **`/simple/`** index — a different CDN path with independent propagation lag — so the existing poll doesn't protect the build.

## Proposed fix

Wrap the `just ci-e2e-cli-build` call in each of the three build steps in an inline bash retry loop (matching the repo's existing inline-retry idiom in `release-cli.yml`, no new third-party action). Buildkit layer caching makes a re-run cheap — only the failed `uv sync` layer re-executes.

```yaml
for i in $(seq 1 3); do
  if just ci-e2e-cli-build "$ECR_URL:jd-<variant>" "$BUILD_ARGS"; then exit 0; fi
  echo "Attempt ${i}/3 failed, retrying in 30s..."
  sleep 30
done
echo "::error::CLI E2E image (<variant>) build failed after 3 attempts"
exit 1
```

`.github/workflows/e2e-cli.yml` — steps `Build CLI E2E image (bare|aws|aws-k8s)`.

Alternative considered: poll the `/simple/` index (not just JSON) before building, but a build-step retry is simpler and also covers transient network/index errors.

### LINKED ISSUE #252: Harden CI: handle stale (undeployed) projects in S3 store gracefully
## Problem

When an EKS project exists in the S3 store but has no live infrastructure (e.g. after a
`destroy-after` run, or manual teardown without store cleanup), two CI jobs crash:

1. **`find-takedown-eks`** (used in `cleanup-existing` step of `e2e-eks-oidc-fresh.yml`):
   restores the project, runs `jd down`, which fails because terraform can't plan — the
   OAuth secret can't be populated (no `secret_arn` output) and variables aren't wired.

2. **`ci-restore-eks`** (used in `eks-functional-test` of `e2e-cli.yml`):
   restores the project, then `jd config --restore-secrets` fails with
   `Cannot restore secret for 'oauth_app_client_secret': output 'secret_arn' not found`.

In both cases the store entry is stale — there's nothing deployed, so destroy is a no-op
and secret restore is impossible.

## Desired behavior

After restoring a project from the store, detect whether the infrastructure is actually
deployed (e.g. terraform state is empty, or key outputs like `secret_arn` are missing).

- **`find-takedown-eks`**: if state is empty, skip `jd down` and go straight to
  `jd projects delete` to clean up the stale store entry.
- **`ci-restore-eks`**: if `--restore-secrets` fails because outputs are missing, surface
  a clear error: "Project exists in store but is not deployed — run the fresh deploy
  workflow to recreate it, or delete the stale entry with `jd projects delete`."

## Context

Triggered by app #5 (`release-eks-oidc` / `tf-aws-eks-oidc-98a125fd`) being in the store
after its cluster was destroyed. Required manual `jd projects delete` to unblock CI.

## Files to change

- `scripts/find_takedown_eks.py` (and likely `find_takedown_base.py`)
- `scripts/ci_restore_eks.py` (and likely `ci_restore_base.py`)

JGuinegagne:
This PR a/ hardens release and image build against `test-pypi` ultimate consistency race conditions, b/ hardens against projects taken down (`state=<empty>`) but kept in store, c/ improve web-app e2e tests

Closes: #251
Closes: #252
Closes: #246

### Implementation (web-app tests)
- create a reusable fixture: `dex_oauth_web_app` which provides an instance of `WebAppNavigator`
- use the `WebAppNavigator` instance to handle most UI flows -> benefit, we only need to update the tests in the plugin and reuse in multiple templates

### Test coverage
- create default workspace, verify transition Starting > Running, verify can open JupyterLab
- can stop, restart own workspace and then open JupyterLab
- can discover other users' workspaces
- can open other users' public workspaces
- cannot open other users' private workspaces

### Testing
- ran `test_web_app` E2E tests

### DIFF
--- a/.github/workflows/e2e-cli.yml
+++ b/.github/workflows/e2e-cli.yml
@@ -64,7 +64,17 @@ jobs:
           BUILD_ARGS="--build-arg INSTALL_MODE=pypi"
           BUILD_ARGS="$BUILD_ARGS --build-arg INSTALL_VARIANT=bare"
           BUILD_ARGS="$BUILD_ARGS --build-arg PKG_VERSION=${{ inputs.pkg-version }}"
-          just ci-e2e-cli-build "$ECR_URL:jd-bare" "$BUILD_ARGS"
+          for i in 1 2; do
+            if just ci-e2e-cli-build "$ECR_URL:jd-bare" "$BUILD_ARGS" 2>&1 | tee /tmp/build.log; then exit 0; fi
+            if ! grep -qi "No solution found\|no version of jupyter-deploy" /tmp/build.log; then
+              echo "::error::Build failed with non-transient error"
+              exit 1
+            fi
+            echo "Attempt ${i}/2 failed (package not yet available), retrying in 60s..."
+            sleep 60
+          done
+          echo "::error::CLI E2E image (bare) build failed after 2 attempts"
+          exit 1
 
       - name: Push to ECR
         run: just ci-e2e-cli-push bare
@@ -113,7 +123,17 @@ jobs:
           BUILD_ARGS="--build-arg INSTALL_MODE=pypi"
           BUILD_ARGS="$BUILD_ARGS --build-arg INSTALL_VARIANT=aws"
           BUILD_ARGS="$BUILD_ARGS --build-arg PKG_VERSION=${{ inputs.pkg-version }}"
-          just ci-e2e-cli-build "$ECR_URL:jd-aws" "$BUILD_ARGS"
+          for i in 1 2; do
+            if just ci-e2e-cli-build "$ECR_URL:jd-aws" "$BUILD_ARGS" 2>&1 | tee /tmp/build.log; then exit 0; fi
+            if ! grep -qi "No solution found\|no version of jupyter-deploy" /tmp/build.log; then
+              echo "::error::Build failed with non-transient error"
+              exit 1
+            fi
+            echo "Attempt ${i}/2 failed (package not yet available), retrying in 60s..."
+            sleep 60
+          done
+          echo "::error::CLI E2E image (aws) build failed after 2 attempts"
+          exit 1
 
       - name: Push to ECR
         run: just ci-e2e-cli-push aws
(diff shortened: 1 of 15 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

bhanderig (approved):

PR #263, opened 2026-06-15
fix: eks-oidc role assumption, merge caller user/role for stability

### LINKED ISSUE #257: Demo: revisit Admin role as default assumption
The `eks-oidc` template assumes that there exists an `Admin` role in the AWS account of the caller of `jd config` / `jd up` and adds it by default to the IAM principals that map to a cluster administrator. This assumption is incorrect: we tried with an external AWS account and such a role does not exist. Instead, it showed up a `<arn>:user/Admin`.

### Resolution steps
- keep the additional roles empty by default, rely on authorizing the caller principal
- ensure that terraform awaits the EKS access point before attempting to create k8s resources via helm

JGuinegagne:
This PR improves the `eks-oidc` template to a/ support user as calling role, as opposed to only role, b/ remove the assumption that an `Admin` role exists in the account, c/ exposes `admin_user_names` variable. Furthermore, it improves the CX by automatically merging the caller user or role into the list of admins..

Closes: #257 

### User experience
- if user just runs `jd config` w/ an IAM role, `terraform` a/ finds the role name, b/ create an EKS entry point for it
- if user just runs `jd config` w/ an IAM user, `terraform` a/ finds the user name, b/ create an EKS entry point for it
- if user runs a/ passes this role/user to `--admin-role-names` or `--admin-user-names` when running `jd config`, then a/later uses a different role/user to update the project, the EKS entry point linked to the role/user in a/ persist
- therefore, by listing all the roles/users at first, users can ensure the state remains stable when they switch between such users/roles and all their roles/users retain cluster admin access

### Implementation
- update the default `admin_role_names`: `["Admin"]` > `[]`
- identify whether caller is an IAM role or an IAM user
- merge the role / user w/ the `admin_role_names` or `admin_user_names`, avoid duplication
- create EKS access entry point for both
- expose `admin_user_names` as a new template variable

### Testing
- updated the template, ran `jd config & jd up`, switched roles, ran `jd config`, observed no state diff.


michaelnchin (approved):
LGTM - thank you @JGuinegagne !

bhanderig (approved):

PR #267, opened 2026-06-16
feat: add roborev review image and IAM template

andrii-i:
Part of jupyter-infra/jupyter-k8s#400. roborev provides the review engine but not the infrastructure to run it. This PR adds that infrastructure for #400: the review runner image, a workflow to build and publish it, and an IAM/ECR template, so reviews run with Claude on Amazon Bedrock and no API keys stored anywhere, following the existing jupyter-deploy pattern next to the e2e image build and CI IAM template. Pairs with jupyter-infra/jupyter-k8s#410, which consumes what this publishes, so deploy and configure this PR first.

Review image (`.github/review/Dockerfile`)
- `node:24-bookworm-slim` with version-pinned roborev, the claude-code agent, and gh.
- The roborev tarball is verified against a pinned SHA256 and fails closed on mismatch.
- Holds no credentials; AWS and GitHub credentials are injected at runtime. The dockerignore is `*`, so the image is self-contained with no build context.

Build workflow (`.github/workflows/review-build-image.yml`)
- Assumes the publish role via OIDC, logs in to ECR, builds the image, and pushes it.
- Publishes occasionally (push to main touching `.github/review/**`, or `workflow_dispatch`), not per PR.
- Inert until `REVIEW_PUBLISH_ROLE_ARN` and `REVIEW_ECR_REPOSITORY` are set: the build job is gated on both, so it can be merged before the infrastructure exists.

IAM/ECR template (`libs/jupyter-infra-tf-aws-review-ci/`)
- New `jd init`/`jd up` template with the same structure as `jupyter-infra-tf-aws-iam-ci`.
- Creates the review ECR repository and two roles: a publish role (ECR push, trust scoped to this repo's `review` GitHub environment) and a run role (ECR pull plus `bedrock:InvokeModel`/`InvokeModelWithResponseStream` on configured model ARNs only, trust scoped to `environment:review` for an explicit `review_repos` list).
- Both roles carry the same deny-self-modify hardening as the existing e2e/release roles, and unlike those attach only their scoped policy, no AdministratorAccess.
- Registered for pytest and mypy alongside the other libs.

#### Note on decoupling the build from state

Unlike `e2e-build-image.yml`, the build does not restore the jd project or read terraform outputs. It only needs the ECR repository URL, which is a template output, so it takes it as the `REVIEW_ECR_REPOSITORY` repo variable. That keeps the publish role to ECR push only, with no access to the project store or terraform state.

### Testing

- `uv run pytest libs/jupyter-infra-tf-aws-review-ci/tests/` -> 26 passed (version, manifest, and `variables.yaml` (schema v2) / `variables.tf` / `defaults-all.tfvars` consistency).
- `terraform validate`, `ruff`, `mypy`, `actionlint`, `yamllint` -> clean.
- `docker build --check` on the review image -> no warnings; the pinned roborev SHA256 matches upstream and `sha256sum --check` fails closed in the base image.

The publish path (OIDC -> ECR push) is inert until the two repo variables are set, so it merges before the infrastructure exists; the first `review-build-image.yml` run during rollout exercises it live.

### Rollout

1. `jd init`/`jd up` the `tf-aws-review-ci` template (set `github_org`, `review_repos`, `bedrock_model_arns`).
2. Create the `review` GitHub Actions environment in this repo and set two repo variables from the template outputs: `REVIEW_PUBLISH_ROLE_ARN` (from `review_publish_iam_role_arn`) and `REVIEW_ECR_REPOSITORY` (from `review_image_repository_url`).
3. Run `review-build-image.yml` to publish the image.
4. Configure jupyter-k8s per #410 (run role ARN, image ref, Bedrock model).


.github/review/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# A standalone runtime for `roborev ci review`: roborev plus the claude-code
+# agent it shells out to, git, gh (roborev posts the PR comment via gh), and jq.
+# Consumer workflows assume an IAM role, pull this image from ECR, and run it
+# against a PR ref range. The image holds no credentials; AWS and GitHub creds
+# are passed in as environment at run time.
+#
+# All tool versions are pinned so a given image tag is reproducible and the
+# supply chain is auditable. The roborev tarball is checksum-verified.
+
+FROM node:22-bookworm-slim
```
JGuinegagne: could we use the latest stable node version (`node:24`) ?

.github/review/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    && gh --version
+
+# claude-code agent — pinned npm release. roborev invokes `claude`.
+RUN npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
+    && claude --version
+
+LABEL org.opencontainers.image.source="https://github.com/jupyter-infra/jupyter-deploy"
+LABEL org.opencontainers.image.description="roborev CI review runner (roborev + claude-code agent)"
+
+CMD ["roborev", "--help"]
```
JGuinegagne, line 59: consider switching user instead of running as root.

.github/workflows/review-build-image.yml
```diff
@@ -0,0 +1,63 @@
...
+#       REVIEW_PUBLISH_ROLE_ARN  <- review_publish_iam_role_arn
+#       REVIEW_ECR_REPOSITORY    <- review_image_repository_url
+#       REVIEW_AWS_REGION        optional, defaults to us-west-2
+
+name: Build review image
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
```
JGuinegagne, line 25: curious: why rebuild + publish on each push to `main` as opposed to a/ manual trigger or b/ when detecting changes, e.g. based on versions or hash of a set of files?

libs/jupyter-infra-tf-aws-iam-review/jupyter_infra_tf_aws_iam_review/template/engine/variables.tf
```diff
@@ -0,0 +1,104 @@
...
+    Bedrock model and inference-profile ARNs the run role may invoke.
+
+    For cross-region inference profiles, include both the inference profile ARN
+    and the underlying foundation model ARNs in each region the profile routes to.
+
+    Example:
+    [
+      "arn:aws:bedrock:us-west-2:123456789012:inference-profile/us.anthropic.claude-sonnet-4-*",
+      "arn:aws:bedrock:us-*::foundation-model/anthropic.claude-*"
+    ]
```
JGuinegagne: could we compile the ARNs from the `region` variable and from the AWS `account` of the caller?

libs/jupyter-infra-tf-aws-review-ci/README.md
```diff
@@ -0,0 +1,69 @@
...
+- GitHub repositories with a `review` Actions environment
+
+## Usage
+This terraform project is meant to be used with the [jupyter-deploy](https://github.com/jupyter-infra/jupyter-deploy/tree/main/libs/jupyter-deploy) CLI.
+
+### Installation (with pip):
+Recommended: create or activate a python virtual environment.
+
+```bash
+pip install jupyter-deploy[aws]
```
JGuinegagne: nit: to support all shell, use `pip install "jupyter-deploy[aws]"`

libs/jupyter-infra-tf-aws-review-ci/README.md
```diff
@@ -0,0 +1,69 @@
...
+pip install jupyter-infra-tf-aws-review-ci
+```
+
+### Project setup
+```bash
+mkdir review-ci
+cd review-ci
+
+jd init . -P aws -I review -T ci
```
JGuinegagne: `-I` stands for infrastructure.

Here you infra is likely `iam` (primarily setup roles), or possibly `ecr`

JGuinegagne (commented):
Looks good overall, thanks Andrii!

I wonder if we adding to the existing `ci` template as opposed to creating a new template would be better.

Not that strong of an opinion, and it might be helpful to have a separate template for repos that do not need the secrets/e2e image ECR infra. However, sleep on this please and consider whether that's the right choice.

Another question: how will we commit the skills / configs that roborev should use? They could either be baked into the image, or pulled from the PR, what's your take?

.github/review/Dockerfile
andrii-i: Updated to using `node:24`

.github/review/Dockerfile
andrii-i: The image now runs as the unprivileged `node` user provided by the `node:24-bookworm-slim` base image. `/home/node` is owned by `node` from that base image, so it's writable for the roborev/claude-code/gh runtime state.

Alternatively a numeric UID could be used, but in this case `USER node` is clearer and there's no benefit to the numeric form here.

.github/workflows/review-build-image.yml
andrii-i: Note `paths` filters at LL26-28. They scope rebuild + publish on pushes that change review image definition or workflow only.
```
    paths:
      - ".github/review/**"
      - ".github/workflows/review-build-image.yml"
 ```

File: libs/jupyter-infra-tf-aws-iam-review/jupyter_infra_tf_aws_iam_review/template/engine/variables.tf (hunk 4 above)
Comment (andrii-i) [reply libs/jupyter-infra-tf-aws-iam-review/jupyter_infra_tf_aws_iam_review/template/engine/variables.tf]: Good idea, implemented deriving the bedrock inference-profile ARNs from the caller's account and the deployment region.

The foundation-model ARNs have no account, since AWS owns them, and a cross-region profile spans several regions that each must be allowed, so those stay an explicit list.

File: libs/jupyter-infra-tf-aws-review-ci/README.md (hunk 6 above)
Comment (andrii-i) [reply libs/jupyter-infra-tf-aws-review-ci/README.md]: Good point. It's `iam` as the template is primarily the roles.

**Comment by andrii-i** [comment]
@JGuinegagne thank you for giving this a look. Added the review resources to `tf-aws-iam-ci` rather than a separate template, to avoid a second GitHub Actions OIDC provider: it's a per-account singleton, so two templates that both declare it collide on deploy and need `create_oidc_provider = false` set on whichever runs second. A separate template would make sense for an account that wants reviews without the secrets/e2e infra, but that's not the case here, so consolidating made more sense.

On baking the configs into the image vs pulling them from the repo: both, split by what each is. roborev and the claude-code agent are baked in at pinned versions, so a given image tag always reviews with the same toolchain. The review config, which review types run and settings like the model, is pulled from the reviewed repo's `.roborev.toml` at review time, so each repo version-controls its own reviews and nothing repo-specific is in the image.

Give this a look. If it all makes sense, I'll deploy `tf-aws-iam-ci` to the Jupyter AWS CI account. I'm updating https://github.com/jupyter-infra/jupyter-k8s/pull/410 in the meantime to work with these changes. If everything looks good here, I'd say we can wrap this all up including https://github.com/jupyter-infra/jupyter-k8s/pull/410 in the next couple of days / until the end of the week.

**Comment by JGuinegagne** [comment]
Security-focused review of the roborev review image + IAM template. The design is sound — scoped OIDC trust, deny-self-modify on both roles, no `AdministratorAccess`, minimal Bedrock action set, and an explicitly stated/bounded prompt-injection blast radius (pull image + spend Claude tokens). roborev is SHA256-verified and fails closed.

**Validated as safe:**
- Bedrock resource scope (`us-*`/`claude-*`) is broad but deliberate; blast radius bounded to token spend, actions minimal — acceptable as a default.
- `dockerignore: *` keeps the build context credential-free; image runs as the unprivileged `node` user.
- Region parsing in the justfile is from a trusted TF-output repo variable and fails cleanly on a malformed URL.

**Findings (4 non-blocking), posted inline:**
1. `[Hardening]` `gh`/`claude-code` aren't checksum-verified; docs overclaim "supply chain is auditable."
2. `[Security]` Trust boundary depends on the `review` GH environment having protection rules — currently undocumented.
3. `[Reliability]` `create_review_resources=true` + default empty `review_repos` → unassumable ghost role; add a plan-time guard.
4. `[Reliability]` `REVIEW_AWS_REGION` default (`us-west-2`) ≠ template `region` default (`us-east-1`); wire it from the output.

Nit: PR body says `node:22-bookworm-slim`; the Dockerfile uses `node:24-bookworm-slim`.

File: .github/review/Dockerfile [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+
+# gh CLI — pinned release tarball.
+RUN curl -sfLO "https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_amd64.tar.gz" \
+    && tar xzf "gh_${GH_VERSION}_linux_amd64.tar.gz" \
+    && install "gh_${GH_VERSION}_linux_amd64/bin/gh" /usr/local/bin/gh \
+    && rm -rf "gh_${GH_VERSION}_linux_amd64" "gh_${GH_VERSION}_linux_amd64.tar.gz" \
+    && gh --version
+
+# claude-code agent — pinned npm release. roborev invokes `claude`.
+RUN npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
```
JGuinegagne, line 46: **[Hardening]** The header (lines 9-10) and the PR body state the "supply chain is auditable / the roborev tarball is checksum-verified," which reads as if all three tools are integrity-checked. Only `roborev` is — `gh` (line 37) and `claude-code` (this line) are version-pinned but not checksum-verified.

That's a defensible trust call (`gh`/`claude-code` come from more-trusted sources than the third-party `roborev-dev` org), but please make it explicit: (1) add a 2-sentence comment near the `gh`/`npm` installs noting they're version-pinned but **not** checksum-verified, and why that's acceptable; and (2) reword the header's "supply chain is auditable" line (and the matching PR-body sentence) so it doesn't imply all three are verified — e.g. "tool versions are pinned for reproducibility; the roborev tarball is additionally SHA256-verified."

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/review.tf
```diff
@@ -0,0 +1,136 @@
...
+  oidc_provider_url  = local.oidc_provider_url
+  github_org         = var.github_org
+  github_repos       = [var.publish_repo]
+  oidc_trust_subject = "environment:review"
+  policy_arns        = [aws_iam_policy.review_publish[0].arn]
+  tags               = local.default_tags
+}
+
+# Run role — assumed by consumer repos to pull the image and run reviews.
+module "role_review_run" {
```
JGuinegagne, line 131: **[Security]** The run role's only trust gate is `sub = repo:<org>/<repo>:environment:review`. OIDC proves a job *declared* the `review` environment — it does **not** prove the workflow file was trusted. If the `review` GitHub Actions environment has no protection rules, a malicious PR (or a `pull_request_target` workflow) in a `review_repos` repo could assume this run role on its own terms, collapsing the boundary this template sets up.

Please add a line to `.github/AGENT.md` (and ideally the rollout steps) documenting that the `review` environment **MUST** carry protection rules — required reviewers and/or restricted branches/tags — for this trust scoping to hold.

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/variables.tf
```diff
@@ -403,3 +403,66 @@ variable "github_oauth_app_client_secret_6" {
...
+  description = <<-EOT
+    GitHub repository that builds and pushes the review image (assumes the
+    publish role). Used only when create_review_resources is true.
+
+    Example: jupyter-deploy
+  EOT
+  type        = string
+}
+
+variable "review_repos" {
```
JGuinegagne, line 429: **[Reliability]** `review_repos` defaults to `[]`, but when `create_review_resources = true` the run role is still created with an empty `StringLike` `sub` list — producing a role nobody can assume. The operator applies cleanly, publishes an image, and only discovers the dead role when a consumer's review job fails to assume it.

Better to fail fast than create a ghost resource. Add a plan-time guard that errors when `create_review_resources && length(review_repos) == 0`. Since this is a cross-variable condition, it needs a `check` block or a resource precondition rather than a `validation {}` on the variable itself, e.g.:

```hcl
check "review_repos_set_when_enabled" {
  assert {
    condition     = !var.create_review_resources || length(var.review_repos) > 0
    error_message = "review_repos must be non-empty when create_review_resources is true."
  }
}
```

.github/workflows/review-build-image.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install just
+        uses: ./.github/actions/install-just
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ vars.REVIEW_PUBLISH_ROLE_ARN }}
+          aws-region: ${{ vars.REVIEW_AWS_REGION || 'us-west-2' }}
```
JGuinegagne: **[Reliability]** `REVIEW_AWS_REGION` defaults to `us-west-2`, but the `tf-aws-iam-ci` template's `region` defaults to `us-east-1` — so out of the box this credentials step targets a different region than where the ECR repo actually lives. It happens to still work because `ci-review-push/pull` re-derive the region from the ECR URL for the login, but the three defaults disagree (`us-west-2` here, `us-east-1` in the template, `us-*` for Bedrock).

Suggest: in rollout step 2, list `REVIEW_AWS_REGION ← region` (from the template's `region` output) alongside the other two repo variables, and/or change this workflow's default to `us-east-1` to match the template default.

JGuinegagne (changes requested):
Any way to reuse the existing `module.iam_role` (possibly adding parameters) instead of adding a `review_role` module?

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/review.tf
andrii-i: Good point. Documented the requirement in `.github/AGENT.md` and the template README (and will of course set the GitHub environment rules at deploy).

After the initial release, as a follow-up, we should add a stronger, template-enforced control: scoping the run-role trust to `job_workflow_ref` of a pinned reusable review workflow, so only that workflow can assume the role regardless of environment config. I'll track this.

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/variables.tf
andrii-i: Agreed on failing fast. A `check` block only warns and lets the apply continue, so it would still create the run role nobody can assume. I used a resource `precondition` on the run policy instead, which errors and stops the apply.

.github/workflows/review-build-image.yml
andrii-i: Aligned the default to `us-east-1` to match the template's `region` default, and listed `REVIEW_AWS_REGION <- region` in the prereq comment alongside the other two repo vars so the three stay consistent.

andrii-i:
> Any way to reuse the existing module.iam_role (possibly adding parameters) instead of adding a review_role module? 

@JGuinegagne Good point. Reused `module.iam_role` for the review roles instead of the separate module.

andrii-i:
@JGuinegagne thank you for the review. Updated everyhting based on the comments.

If everything looks good, after this merges I'll deploy `tf-aws-iam-ci` to the Jupyter AWS CI account, then verify https://github.com/jupyter-infra/jupyter-k8s/pull/410 against it.

.github/review/Dockerfile
JGuinegagne, line 23: non-blocking: might need `apt-get upgrade` as well to pull the latest deps

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/review.tf
```diff
@@ -103,12 +103,19 @@ resource "aws_iam_policy" "review_run" {
   })
 
   tags = local.default_tags
+
+  lifecycle {
+    precondition {
+      condition     = length(var.review_repos) > 0
```
JGuinegagne, line 109: sanity-check: shouldn't this apply only when the review flag is `true` ?

JGuinegagne (approved):

.github/review/Dockerfile
andrii-i: I think we should definitely upgrade packages. We can do that with `apt-get upgrade` or by pulling the latest base image. Either way, it should run on a schedule (say weekly), since this workflow currently only fires on a push that changes `.github/review/**` or the workflow file (or a manual run), which might be rare. I'll track it as a follow-up.

libs/jupyter-infra-tf-aws-iam-ci/jupyter_infra_tf_aws_iam_ci/template/engine/review.tf
andrii-i: It should, and it is: the precondition is attached to `aws_iam_policy.review_run`, which is only created when `create_review_resources` is true.

Issue #270, opened 2026-06-19
EKS OIDC: orphaned WorkspaceAccessStrategy breaks workspace creation


JGuinegagne:
# Issue: orphaned WorkspaceAccessStrategy breaks workspace creation

## Symptom
UI workspace creation fails with a misleading `403 Forbidden`. Real cause in web-app logs:
`admission webhook "mworkspace-v1alpha1.kb.io" denied the request: failed to add
finalizer to AccessStrategy: referenced AccessStrategy oauth-access-strategy not found
in namespace jupyter-k8s-shared`. Not an RBAC/team-permission problem.

## Root cause
`oauth-access-strategy` (a `WorkspaceAccessStrategy` CR) is **owned by the router release**
(`jupyter-k8s-aws-oidc`), but its **CRD lifecycle is owned by the operator release**
(`jupyter-k8s`). Tearing down / replacing the operator deletes the CRD, which deletes every
CR of that kind — including `oauth-access-strategy` — even though a different release owns it.
The router release is not re-run, so the CR is never recreated.

The Terraform helm provider only diffs a release's chart+values, not its in-cluster objects.
So after the CR is orphaned, `jd config` / `jd up` both report **"No changes"** and never
self-heal. `helm.sh/resource-policy: keep` on the CRD protects normal upgrades but NOT an
explicit replace/uninstall of the operator.

## Done
- Live fix (sandbox-eks): re-applied the CR from `helm get manifest jupyter-k8s-aws-oidc`.
- Template fix: added `kubernetes_namespace_v1.shared` to `helm_release.workspace_router`
  `depends_on` (engine/helm.tf) — fixes the fresh-deploy ordering race only.

## Still open
- The `depends_on` edge does NOT protect against operator-replace orphaning the CR.
- Need a durable guard, e.g.:
  - `jd health --components` assertion that the strategy exists, or
  - decouple the strategy CR from operator-CRD lifecycle, or detect/repair in-cluster drift.

JGuinegagne:
## Proposal: detect + self-heal orphaned CRs in `jd config` / `jd up`

I dug into the operator code (`../jk8s`) to nail down the mechanism before proposing a fix. Two findings change the picture.

### Finding 1 — in the deployed operator, the AccessStrategy was *unprotected*

[...]

PR #271, opened 2026-06-19
feat: image cmds

JGuinegagne:
This PR adds `jd image` commands to provide visibility to the admins on the application images they are using. The new command supports discovering images and retrieving their vulnerability scan results.

Also in this PR:
- upgrade `base` template `jupyterlab` and `log-rotator` images
- upgrade `eks-oidc` template `jupyterlab` image
- partially paper-cut of dependencies chain in `eks-oidc` - more details for deeper issue in: #270 

Next steps:
- add image scan status to `jd health`
- think more on exploitability, what to display to admins to make informed decision, possibly tolerate CVEs
- provide cmd to review details of a specific CVE
- add documentation on how to enable [inspectorv2](https://docs.aws.amazon.com/inspector/v2/APIReference/Welcome.html) scans for ECR images in the account

### User experience
- `jd image list` --> list application images for a project
- `jd image show --name <>` --> details about repository location, tags, and scanning status of an image
- `jd image show` --> convenience if the template only has 1 application image
- `jd image tags [--name <>]` --> list the image tags
- `jd image vulnerabilities [--name <> --tag <>]` --> list the CVEs in the image detected by the scanner
- `list`, `tags` support `--text` and `--json` (for agent's use)
- `show`, `vulnerabilities` support `--json` (for agent's use)

### Implementation
- add `api/aws/ecr` to describe ECR repository and basic ECR scan result (on image push)
- add `api/aws/inspectorv2` to detect whether inspectorv2 is enabled for the account, and retrieve scanning results
- add command runners for `ecr` and `inspectorv2`, register in aws command runner
- automatically detect if `inspector-v2` is enabled for ECR images in the account, use reports it generate over basic ECR scan if that's the case
- add `ImageHandler` to `<jd>/handlers/resources`
- add `image` commands group to `<jd>/cli`
- add doc references for `image` commands
- add missing `apt-get upgrade -y` to `dockerfile` of `jupyterlab`/`log-rotator` image in `base` template
- add missing `apt-get upgrade -y` to `dockerfile` of `jupyterlab` image in `eks-oidc` template
- build `jupyterlab` images for `base` and `eks-oidc` templates of `py3.13-bookworm-slim` debian image
- make `router` and `workspace_defaults` chart of `eks-oidc` template depend on `jupyter-k8s` chart and `shared` namespace

### Testing
- add E2E tests for image commands to `eks-oidc` tests, mark as CLI so they run in the functional tests of the CLI
- update `eks-oidc` image, redeploy, test manually

### Screenshots

**jd image --help**
Screenshot 2026-06-19 at 7 23 06 AM

**jd image list**
Screenshot 2026-06-19 at 7 22 26 AM

**jd image show**
Screenshot 2026-06-19 at 7 23 58 AM

**jd image tags**
Screenshot 2026-06-19 at 7 24 52 AM

**jd image vulnerabilities**
Screenshot 2026-06-19 at 7 25 52 AM


libs/jupyter-deploy/jupyter_deploy/api/aws/sts/sts_identity.py
```diff
@@ -17,6 +17,11 @@ def get_caller_arn(sts_client: STSClient) -> str:
     return identity["Arn"]
 
 
+def get_caller_account(sts_client: STSClient) -> str:
+    """Return the AWS account ID of the caller's identity via STS:GetCallerIdentity."""
+    return get_caller_arn(sts_client).split(":")[4]
```
bhanderig: `[Hardening]` `GetCallerIdentity` returns `Account` as a top-level field in the response. Parsing it from the ARN works but is strictly more fragile than reading the direct field.

The partition split in `get_partition` is justified (no `Partition` field exists), but for account ID the response hands it to you:

```python
def get_caller_account(sts_client: STSClient) -> str:
    """Return the AWS account ID of the caller's identity via STS:GetCallerIdentity."""
    identity = sts_client.get_caller_identity()
    return identity["Account"]
```

Avoids depending on the colon-delimited ARN format assumption and the extra indirection through `get_caller_arn`.

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_inspector_runner.py
```diff
@@ -0,0 +1,241 @@
...
+        self.region_name = region_name
+        self.inspector_client: Inspector2Client = boto3.client("inspector2", region_name=region_name)
+        self.ecr_client: ECRClient = boto3.client("ecr", region_name=region_name)
+        self._sts_client: STSClient = boto3.client("sts", region_name=region_name)
+        self._inspector_enabled: bool | None = None
+
+    def _is_inspector_enabled(self) -> bool:
+        if self._inspector_enabled is not None:
+            return self._inspector_enabled
+        self._inspector_enabled = inspector2_findings.is_ecr_scanning_enabled(self.inspector_client, self._sts_client)
```
bhanderig, line 51: `[Reliability]` If the caller's IAM role lacks `inspector2:BatchGetAccountStatus`, this throws an unhandled `ClientError(AccessDeniedException)` — crashing the command with a raw boto stack trace.

Many users will have ECR permissions without Inspector2 (it's a separate service requiring explicit enablement). Consider catching `ClientError` and treating permission errors as "not enabled" with a warning:

```python
def _is_inspector_enabled(self) -> bool:
    if self._inspector_enabled is not None:
        return self._inspector_enabled
    try:
        self._inspector_enabled = inspector2_findings.is_ecr_scanning_enabled(
            self.inspector_client, self._sts_client
        )
    except botocore.exceptions.ClientError:
        self.display_manager.warning(
            "Inspector2 not accessible, falling back to ECR basic scan"
        )
        self._inspector_enabled = False
    return self._inspector_enabled
```

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_inspector_runner.py
```diff
@@ -0,0 +1,241 @@
...
+            result = self._list_vulnerabilities_inspector(repository_name, image_tag)
+            if json.loads(result["Vulnerabilities"].value):
+                return result
+            # Inspector returned empty — validate the tag exists via ECR
+            self.display_manager.info(f"No Inspector findings, verifying tag exists: {repository_name}:{image_tag}")
+
+        try:
+            self.display_manager.info(f"Reading ECR scan findings for {repository_name}:{image_tag}")
+            return self._list_vulnerabilities_ecr_basic(repository_name, image_tag)
+        except self.ecr_client.exceptions.ImageNotFoundException:
```
bhanderig, line 208: `[Reliability]` `describe_image_scan_findings` can also throw `ScanNotFoundException` when the image exists but has never been scanned (e.g., scan-on-push disabled, or image predates scan configuration). This would surface as a raw boto exception to the user.

Consider catching it alongside `ImageNotFoundException`:

```python
try:
    ...
    return self._list_vulnerabilities_ecr_basic(repository_name, image_tag)
except self.ecr_client.exceptions.ImageNotFoundException:
    raise ImageTagNotFoundError(repository_name, image_tag) from None
except self.ecr_client.exceptions.ScanNotFoundException:
    # Image exists but was never scanned
    return self._empty_result("ECR Basic")
```

Or return a clear message: "No scan results — enable scan-on-push or trigger a manual scan."

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_inspector_runner.py
```diff
@@ -0,0 +1,241 @@
...
+    ) -> dict[str, ResolvedInstructionResult]:
+        repository_name_arg = require_arg(resolved_arguments, "repository_name", StrResolvedInstructionArgument)
+        image_tag_arg = require_arg(resolved_arguments, "image_tag", StrResolvedInstructionArgument)
+        repository_name = repository_name_arg.value
+        image_tag = image_tag_arg.value
+
+        if self._is_inspector_enabled():
+            self.display_manager.info(f"Reading Inspector findings for {repository_name}:{image_tag}")
+            result = self._list_vulnerabilities_inspector(repository_name, image_tag)
+            if json.loads(result["Vulnerabilities"].value):
```
bhanderig: `[Reliability]` When Inspector is enabled and returns zero findings, this falls through to ECR basic scan and **returns those results**. If ECR basic has stale push-time findings that Inspector has since cleared (continuous re-scan), the user sees outdated CVEs from the inferior scanner.

If the intent is only to validate the tag exists when Inspector returns empty, consider:
1. Return the empty Inspector result directly (the authoritative "clean" signal)
2. Validate tag existence as a separate check that doesn't replace the result

ECR basic should be the data source only when Inspector is *disabled*, not when it reports zero findings.

libs/jupyter-deploy-tf-aws-ec2-base/jupyter_deploy_tf_aws_ec2_base/template/services/logrotator/dockerfile.logrotator
```diff
@@ -1,6 +1,6 @@
 FROM alpine:latest
```
bhanderig, line 1: `[Hardening]` (Non-blocking) The other Dockerfiles in this PR pin to `python3.13-bookworm-slim`, but logrotator still uses `alpine:latest`. Since this PR is explicitly improving image security posture (adding vuln scanning, upgrading base images), it'd be a natural time to pin to a specific version (e.g., `alpine:3.21`) for build reproducibility.

JGuinegagne: I will pass on that one for now.

We should pin all the deps for the EC2 base template, but that's a wider story.

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_inspector_runner.py
JGuinegagne: I'll double check but we should catch the standard boto3 error in [this context manager](https://github.com/jupyter-infra/jupyter-deploy/blob/5134f05e971479e8e98a4b27b543280dbcd458c6/libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_error_handler.py#L63)

JGuinegagne:
Full E2E run confirms no regression on base template (at least based on E2E test coverage): https://github.com/jupyter-infra/jupyter-deploy/actions/runs/27825658112

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_inspector_runner.py
JGuinegagne: Confirmed that an AccessDenied error would be caught in the decorator.

bhanderig (approved):

michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #285, opened 2026-06-26
fix: persist project to store even on up failure

### LINKED ISSUE #284: Failed jd up/down does not register the project in the store (orphans cloud resources)
## Problem

When `jd up` (or `jd down`) fails during the terraform apply/destroy, the project
is **never pushed to the remote store** — even though real cloud resources were
created. The deployment becomes an orphan: live in AWS, invisible to the store,
and uncleanable by `find_takedown_*` (which discovers projects via the store).

Observed 2026-06-26 (canary run 28207071874): a fresh eks-oidc deploy failed at
`wait_for_routing_ready` (LE cert issuance slower than the 600s waiter). The EKS
cluster `3e4ec218` was created and is healthy, but no store entry exists, so
`ci-restore-eks 6` / the next canary cleanup can't find or tear it down.

## Root cause

In `cli/app.py`, both `up()` and `down()` call `handler.push_to_store()` only
*after* a successful apply/destroy. A terraform failure raises
`SupervisedExecutionError`, which skips the push and propagates to
`handle_cli_errors` → `typer.Exit`. The local `try` only catches `LogCleanupError`.

## Why the fix is safe

- The store bucket is created on-demand by the CLI (`push_to_store` →
  `find_store`/`ensure_store`), independent of the deploy — no dependency on the
  failed apply.
- `deployment_id` (the store key) is `random_id.postfix.hex`, created at the very
  start of apply, so it's resolvable for essentially any post-start failure.

## Proposed fix

In `up()` and `down()`: capture `SupervisedExecutionError`, run `push_to_store()`
regardless (warn, don't fail, on `ProjectIdNotAvailableError`/`ProjectStoreNotFoundError`),
then re-raise the original error so the exit code/UX is unchanged.

## Acceptance

- Failed `jd up`/`jd down` still registers the project in the store.
- Apply/destroy failure still surfaces with its original exit code.
- Unit tests cover the failure-path push (update any test asserting push is
  skipped on failure).

JGuinegagne:
This PR fixes the `jd up` logic to record the state to the remote store even if `jd up` fails. That way, if the deployment fails in an ephemeral sandbox (e.g. a GitHub job or a lambda function), the terraform state is not lost, no cloud resources are orphaned and the project can be restored later using the standard mechanism.

Closes: #284 

### Implementation
- raise a special `LocalStateNotPersistedError` in `UpHandler` if a/ `apply` fails and b/ the backend store is NOT configured
- catch that error in `<jd>/cli/app:up()`, persist local state to store in that case

### Testing
- forced fail `jd up` using an IAM role w/ missing permission, verified that `jd` saved the state to the remote store
- deleted the local dir
- restored the project from the store
- switched role and deployed

### DIFF
--- a/libs/jupyter-deploy/jupyter_deploy/cli/app.py
+++ b/libs/jupyter-deploy/jupyter_deploy/cli/app.py
@@ -26,9 +26,18 @@ from jupyter_deploy.cli.teams_app import teams_app
 from jupyter_deploy.cli.users_app import users_app
 from jupyter_deploy.cli.variables_decorator import with_project_variables
 from jupyter_deploy.engine.enum import EngineType
+from jupyter_deploy.engine.supervised_execution import DisplayManager
 from jupyter_deploy.engine.vardefs import TemplateVariableDefinition
 from jupyter_deploy.enum import StoreType
-from jupyter_deploy.exceptions import LogCleanupError, OpenWebBrowserError, UrlNotAvailableError
+from jupyter_deploy.exceptions import (
+    LocalStateNotPersistedError,
+    LogCleanupError,
+    OpenWebBrowserError,
+    ProjectIdNotAvailableError,
+    ProjectStoreAccessConfigurationError,
+    ProjectStoreNotFoundError,
+    UrlNotAvailableError,
+)
 from jupyter_deploy.handlers.init_handler import InitHandler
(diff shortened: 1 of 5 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #291, opened 2026-06-30
fix: network probe flakiness

### LINKED ISSUE #290: Harden NetworkPolicy probe helper against pod-name collisions
## Summary

The EKS routing NetworkPolicy E2E tests (`test_routing_network_security.py`)
intermittently fail at probe-pod *creation*, not on the allow/deny assertion:

    Error from server (AlreadyExists): object is being deleted:
    pods "netpol-probe-0" already exists

The probe helper (`pytest_jupyter_deploy/workspaces/network_probe.py`) reuses
fixed pod names and deletes them without waiting, so a probe pod from an earlier
test can still be terminating when a later test tries to recreate the same name.

This flaked the eks-oidc 0.1.1 release gate (2 failures / 132):
https://github.com/jupyter-infra/jupyter-deploy/actions/runs/28461838802/job/84357749338
A local re-run against the same deployment passed 14/14, confirming it is a
test-harness race, not a product or NetworkPolicy defect.

## What to fix

- Make probe-pod identities unique per probe so concurrent/sequential probes
  never collide on a name.
- Ensure a probe never fails because a prior pod is mid-deletion (wait for the
  old pod to be gone, or tolerate the "being deleted" state on create).

## Acceptance

- The routing suite runs repeatedly with no `AlreadyExists` creation failures.
- Real allow/deny assertions are unaffected.

Found during the eks-oidc 0.1.1 release; gates pull the plugin from source, so a
fix on `main` takes effect without republishing pytest-jupyter-deploy.

JGuinegagne:
This PR addresses flakiness in the network policy end-to-end tests by hardening the plugin utilities to use different pod names.

Closes: #290 

### Testing
- ran the routing network E2E tests 5 times in a row
- ran the workspace E2E tests

### DIFF
--- a/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/workspaces/network_probe.py
+++ b/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/workspaces/network_probe.py
@@ -15,18 +15,32 @@ CNI addon must set enableNetworkPolicy=true) — otherwise every probe is allowe
 import json
 import subprocess
 import time
+import uuid
 
 # Operator-generated Service name for a workspace: "workspace-<name>-service".
 _WORKSPACE_SERVICE_FMT = "workspace-{name}-service"
 _CURL_IMAGE = "curlimages/curl:latest"
 # curl exit 28 == operation timeout → the connection was blocked (denied).
 _CURL_TIMEOUT_EXIT = 28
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
LGTM!

bhanderig (approved):

PR #293, opened 2026-07-01
fix: eks-oidc template dependency graph

JGuinegagne:
This PR fixes the resource dependency graph for the `eks-oidc` template such that `jd down` reliably succeeds.

Also in this PR:
- harden home volume E2E test to retry flaky pod-exec (when container is not fully stable)
- bump eks-oidc template version `0.1.1` > `0.1.2`; needed for stuck [release](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/28474410753)
- bump plugin version `0.2.6` > `0.2.7`

### Implementation
- order resource logically 1) {VPC + IAM roles/policies} > 2) Cluster > 3) daemonset AddOns > 4) Managed Nodes Groups > 5) pod-based AddOns > 6) external helm charts (traefik, operator, router) > 7) template charts (workspace defaults, GH RBAC)
- S3, ECR, CodeBuild and EKS-OIDC kept as leaf resources (created / deleted in parallel)
- add `null_resource` to group 3) and 5)
- make node groups (4) depend on 3)
- make external helm charts (6) depend on 3) & 5) (in addition to 4))
- keep template charts (7) to depend on 7)
- this ensures that workspace defaults get destroyed first, then operator/router, then nodes, then cluster, and finally VPC/IAM resources; previously, some resources in 1) and 7) were sibblings, leading to race condition and un-deletable resources in etcd due to webhook.  
- add retry util in plugin, use in flaky e2e test

### Testing
- redeploy `eks-oidc` template, ran full E2E test suite
- ran home volume E2E test suite 5x in a row
- ran `jd down`, verify that the resources delete in the right order

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/eks_addons.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/eks_addons.tf
@@ -3,6 +3,46 @@ resource "time_sleep" "wait_for_nodes" {
   depends_on      = [module.eks_cluster]
 }
 
+# --- Addon ordering aggregators ---
+#
+# These two null_resources are single-source-of-truth barriers that let node
+# groups and Helm releases order themselves against the addons WITHOUT each one
+# re-deriving (and eventually getting wrong) the per-resource addon matrix.
+# No addon depends on a node group or a Helm release, so these introduce no cycle.
+
+# DaemonSet addons a node needs to be functional: vpc-cni (pod IPs) and kube-proxy
+# (Service/ClusterIP routing). The node group depends_on this, so on create the
+# CNI is in place before nodes join, and on destroy the nodes drain BEFORE these
+# are removed. Only DaemonSets belong here: they report healthy with zero nodes,
+# so requiring them before the node group never deadlocks. Deployment addons
+# (coredns, ebs-csi, ...) must NOT be here — they need a schedulable node, so
+# gating the node group on them would be a create-time cycle.
+resource "null_resource" "core_node_addons" {
+  depends_on = [
+    aws_eks_addon.vpc_cni,
+    aws_eks_addon.kube_proxy,
+  ]
+}
+
+# Every cluster addon. The Helm releases (helm.tf) depend_on this, so on create
+# all addons are up before any chart installs, and on destroy every chart
+# uninstalls BEFORE any addon is removed — i.e. the addons a chart relies on
+# (ebs-csi for PVC/PV teardown, coredns for in-cluster DNS, cert-manager,
+# external-dns, ...) stay alive for the entire uninstall. Add a new addon here
+# once and all charts inherit the ordering. external_dns is count-gated; the
+# splat reference resolves to an empty list when disabled.
+resource "null_resource" "cluster_addons" {
+  depends_on = [
+    aws_eks_addon.vpc_cni,
+    aws_eks_addon.kube_proxy,
+    aws_eks_addon.coredns,
+    aws_eks_addon.pod_identity_agent,
+    aws_eks_addon.ebs_csi_driver,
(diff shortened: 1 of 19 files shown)


libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/vpc/outputs.tf
```diff
@@ -6,10 +6,40 @@ output "igw_id" {
...
+#
+# depends_on on the subnet outputs pulls the whole networking set into every
+# consumer's dependency closure:
+#   - on create: routing exists before nodes/ELBs come up;
+#   - on destroy: nodes/cluster/ELBs are torn down before routing is removed.
+# (The IGW also has its own NLB-cleanup ordering via the igw_id output; this is
+# the complementary guarantee for subnet-consuming resources.)
 output "public_subnet_ids" {
   value = aws_subnet.public[*].id
+  depends_on = [
```
bhanderig, line 25: `[Efficiency]` Both subnet outputs share identical `depends_on` lists, meaning a resource consuming only private subnets is also gated on public route table associations (and vice versa). Given that EKS and node groups consume both, this is fine in practice — but if a future resource only needs one subnet type, the cross-coupling adds unnecessary ordering constraints on destroy. Consider splitting if that case arises; fine to leave as-is for now.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/eks_addons.tf
```diff
@@ -3,6 +3,46 @@ resource "time_sleep" "wait_for_nodes" {
...
+}
+
+# Every cluster addon. The Helm releases (helm.tf) depend_on this, so on create
+# all addons are up before any chart installs, and on destroy every chart
+# uninstalls BEFORE any addon is removed — i.e. the addons a chart relies on
+# (ebs-csi for PVC/PV teardown, coredns for in-cluster DNS, cert-manager,
+# external-dns, ...) stay alive for the entire uninstall. Add a new addon here
+# once and all charts inherit the ordering. external_dns is count-gated; the
+# splat reference resolves to an empty list when disabled.
+resource "null_resource" "cluster_addons" {
```
bhanderig, line 34: `[Hardening]` If someone adds a new `aws_eks_addon` to this file but forgets to add it to `cluster_addons.depends_on`, the Helm chart destroy ordering silently regresses. The comments are clear, but a mechanical guard (e.g., a CI grep check like "all aws_eks_addon resources in eks_addons.tf must appear in null_resource.cluster_addons") would catch drift. Low-priority — the AGENT.md documents this, but a lint outlives tribal knowledge.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/vpc/outputs.tf
JGuinegagne: I actually disagree with this. It's more important to keep the dependency graph simple and clear (VPC/IAM > cluster > ...) rather than trying to get marginal optimizations on up/down latency.

The risk of introducing regression (where `jd down` fails) is much more damaging, and it's simpler to say: we will keep the VPC/IAM scaffolding up until the previous stage (e.g. cluster) is fully down.

bhanderig (approved):

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/eks_addons.tf
JGuinegagne: added a unit test!

Issue #71, opened 2025-07-31
Host stop, start and restart commands


JGuinegagne:
Add the following commands and connect it to the base template:
1. `jd host stop`
2. `jd host start`
3. `jd host restart`
When a host is stopped, customers should not incur cloud cost.

### Problem statement
The TLS certificates survive an instance swap/restart because they are saved on the root volume which persists. If a user does `jd down` however, they delete the root volume, and loose their certificates. Let's encrypt only allows to issue 5 certificates per DNS on any 7 day period which is an issue. Users need to be able to stop their instance to no longer incur costs, and restart it without loosing the certificate.

PR #316, opened 2026-07-17
feat: Karpenter autoscaling for eks-oidc template

bhanderig:
## Summary

Replaces the static workspace node groups with a 3-node-group autoscaling architecture for the \`eks-oidc\` template:

- **Platform MNG** — static nodes, Cluster Autoscaler-managed (unchanged)
- **Routing NodePool** — Karpenter-managed, KEDA-scaled on \`traefik_open_connections\`
- **Workspace NodePool(s)** — Karpenter-managed, scale-to-zero when no workspaces running

### Infrastructure changes
- Karpenter: IAM via Pod Identity, SQS interruption queue, Helm release v1.3.3, restart null_resource to fix Pod Identity startup race
- KEDA: Helm release v2.16.1, ScaledObjects for traefik/authmiddleware/web-app
- Prometheus: Helm release v27.0.0 on platform nodes, feeds KEDA triggers
- \`karpenter-nodepools\` Helm chart iterates over \`workspace_nodepools\` list — adding a GPU pool = one new list entry, no new variables
- Remove static workspace node_group module
- Tag EKS cluster SG (not a custom SG) for Karpenter discovery — matches inference-clusters pattern
- \`karpenter.sh/discovery\` subnet tag baked into VPC module
- \`workspace_router\` toleration injected via helm.tf (chart stays template-agnostic)

### New admin variables
- \`workspace_nodepools\` — \`list(map(string))\` with \`name\`, \`instance_families\` (comma-separated), \`disk_size_gb\`, \`max_cpu\`, \`max_memory\` per pool
- \`routing_max_cpu\`, \`routing_max_memory\`, \`routing_instance_categories\`, \`routing_instance_generation_min\`, \`routing_disk_size_gb\`
- \`node_expire_after\`, \`platform_instance_types\` (list)

### CLI changes
- \`jd pool list\` / \`jd pool show --name <pool>\` / \`jd pool status --name <pool>\` — new top-level command backed by dedicated \`PoolHandler\`
- \`jd host list --query\` — free-form label selector, manifest-driven (template-specific labels stay in manifest, not CLI)
- Pool methods moved from \`ClusterHandler\` → dedicated \`PoolHandler\` class

### KEDA scaling signal
All three routing tier services scale on \`sum(traefik_open_connections{entrypoint="websecure"})\`. KEDA AverageValue divides total by replica count before comparing to threshold. Thresholds: traefik=100, authmiddleware=150, web-app=200.

### Reliability fixes (caught during review)
- \`node_role_arn\` was wired to platform MNG role — fixed to Karpenter node role (blocked \`ec2:RunInstances\`)
- All \`local-exec\` scripts write kubeconfig to temp file, preventing corrupt \`~/.kube/config\` from blocking \`jd up\`
- \`platform_max_size >= platform_min_size\` cross-variable validation as Terraform \`check\` block

### Tests
- Unit: 2379 passed
- E2E: \`test_autoscaling.py\` — KEDA ScaledObjects, workspace scale-to-zero, routing NodePool ballast scale-up/down (all passed on live cluster)
- E2E: \`test_pool.py\` — \`jd pool\` command coverage
- E2E: \`test_host.py\` — \`--query\` filter tests

## Test plan
- [x] Unit tests: 2379 passed
- [x] E2E full suite against live EKS cluster — pool, host, autoscaling, health all green
- [x] Fresh \`jd init\` + \`jd config\` + \`jd up\` — cluster up, health checks green
- [x] Workspace create → Karpenter workspace-cpu node → delete → scale-to-zero
- [x] Routing NodePool ballast test — provisions extra node, consolidates after ballast removal
- [x] KEDA ScaledObjects Ready=True, Prometheus connected


docs/source/reference/resource/host.md
```diff
@@ -160,6 +162,7 @@ $ jd host list [OPTIONS]
 **Options**:
 
 * `-p, --path <path>`: Directory of the project whose hosts to list.
+* `--role <str>`: Filter by node role: platform, routing, or workspaces.
```
JGuinegagne: Either we use an enum, or we use free text and we don't provide values like this in the CLI.

Beside, these values are way too template-specific

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
```diff
@@ -0,0 +1,73 @@
...
+      ebs:
+        volumeSize: 50Gi
+        volumeType: gp3
+        encrypted: true
+        deleteOnTermination: true
+---
+apiVersion: karpenter.k8s.aws/v1
+kind: EC2NodeClass
+metadata:
+  name: workspaces
```
JGuinegagne: optional: `workspace-cpu`?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
```diff
@@ -0,0 +1,73 @@
...
+    # code cannot assume the node role. Node agents (CNI, EBS CSI) use hostNetwork
+    # and reach IMDS at hop 1, so they are unaffected.
+    httpPutResponseHopLimit: 1
+    httpEndpoint: enabled
+    httpTokens: required
+    httpProtocolIPv6: disabled
+  blockDeviceMappings:
+    - deviceName: /dev/xvda
+      ebs:
+        volumeSize: 100Gi
```
JGuinegagne: Is that the instance root volume? If so, that seems high for a CPU node, doesn't it?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - key: karpenter.k8s.aws/instance-size
+          operator: In
+          values: ["large", "xlarge", "2xlarge"]
+        - key: karpenter.sh/capacity-type
+          operator: In
+          values: ["on-demand"]
+        - key: kubernetes.io/arch
+          operator: In
+          values: ["amd64"]
+      expireAfter: 504h
```
JGuinegagne: We may want to expose this.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -0,0 +1,91 @@
...
+        kind: EC2NodeClass
+        name: routing
+      taints:
+        - key: jupyter-deploy/role
+          value: routing
+          effect: NoSchedule
+      requirements:
+        - key: karpenter.k8s.aws/instance-family
+          operator: In
+          values: ["c6i", "m6i"]
```
JGuinegagne: Curious, what's the rational for `c` or `m` choices?

Consider exposing this as a JD variable, this seems like an important parameter to expose.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/values.yaml
```diff
@@ -0,0 +1,11 @@
+# Karpenter NodePools and EC2NodeClasses for jupyter-deploy
+# All values are injected by helm.tf — do not set manually.
+
+clusterName: ""
+nodeInstanceProfile: ""
+routingLimitsCpu: "32"
+routingLimitsMemory: "128Gi"
+workspaceLimitsCpu: "512"
+workspaceLimitsMemory: "2048Gi"
```
JGuinegagne: 2TB memory? I don't believe `m` nodes can go this high, that would call for `x` instances I think.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/values.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Comma-separated list of EC2 instance families for workspace nodes
+workspaceCpuInstanceFamilies: "c6i,m6i,r6i"
```
JGuinegagne: shouldn't we allow `7` generation as well?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
```diff
@@ -0,0 +1,211 @@
...
+# Karpenter-provisioned nodes need a security group that allows:
+# - inbound from the cluster security group (kubelet, metrics)
+# - outbound to the cluster API server
+# - inter-node communication for VPC CNI pod networking
+
+data "aws_eks_cluster" "this" {
+  name = var.cluster_name
+}
+
+resource "aws_security_group" "karpenter_nodes" {
```
JGuinegagne: I think this should define this in the `vpc` module.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
```diff
@@ -0,0 +1,211 @@
...
+  to_port                  = 65535
+  protocol                 = "-1"
+  source_security_group_id = aws_security_group.karpenter_nodes.id
+  security_group_id        = data.aws_eks_cluster.this.vpc_config[0].cluster_security_group_id
+  description              = "Allow traffic from Karpenter nodes to cluster"
+}
+
+# ── Karpenter Helm release ────────────────────────────────────────────────────
+
+resource "helm_release" "karpenter" {
```
JGuinegagne: right, I wouldn't use a module for this. It's a deployment singleton and it depends on an external helm chart.

For consistency w/ `logging` and `autoscaler` (similar pattern), how about `platform_karpenter.tf`?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_keda.tf
```diff
@@ -0,0 +1,44 @@
+resource "helm_release" "keda" {
```
JGuinegagne, line 7: same comment, move to `platform_keda.tf` please.

It matters because `modules` are meant to reference either reusable AWS resources (e.g. an IAM role) or core AWS infra (e.g. the cluster, managed node groups, etc). In matters for create/destroy sequence: all this core infra needs to be up before we can start using `helm` to customize the cluster.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_prometheus.tf
```diff
@@ -0,0 +1,78 @@
+resource "helm_release" "prometheus" {
```
JGuinegagne, line 8: same comment

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
```diff
@@ -32,25 +32,25 @@ workspace_app_jupyterlab_app_type        = "jupyterlab"
...
-    role          = "workspaces"
-    instance_type = "c5.2xlarge"
-    ami_type      = "default"
-    disk_size_gb  = "50"
-    min_size      = "2"
-    max_size      = "5"
-    desired_size  = "2"
-  }
-]
+platform_instance_type = "m5.large"
```
JGuinegagne: since we are changing the variable shape, how about making this a list?

MNG can support list of instances, see [API](https://docs.aws.amazon.com/cli/latest/reference/eks/create-nodegroup.html)

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
```diff
@@ -32,25 +32,19 @@ workspace_app_jupyterlab_app_type        = "jupyterlab"
...
+keda_version       = "2.16.1"
+prometheus_version = "27.0.0"
+
+routing_max_cpu    = "32"
+routing_max_memory = "128Gi"
+
+workspace_max_cpu    = "512"
+workspace_max_memory = "2048Gi"
+
+workspace_cpu_instance_families = ["c6i", "m6i", "r6i"]
```
JGuinegagne: Non-blocking: How about using a `list[dict[str,str]]]` for `workspace_nodepools` ?
Something like:
```yaml
workspace-nodepools:
  - name: workspace-cpu-pool
    max_cpu: "96"
    max_memory: "78Gi"
    instance_families: "c6i,m6i,r6i"
    extra-labels: "regular-cpu"  # labels node w/ jupyter-deploy/workspace-category: regular-cpu
    extra-taints-noschedule: "regular-cpu"  # add no-schedule taint on regular-cpu
```

That way, we can support several categories of CPU or GPU workspaces in the future.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/karpenter.tf
```diff
@@ -0,0 +1,115 @@
+# ── Karpenter + KEDA modules ──────────────────────────────────────────────────
```
JGuinegagne: as commented earlier, would it be okay tor reorganize as:
- `platform_karpenter.tf`
- `platform_keda.tf`
- `platform_prometheus.tf`
and remove these modules.

libs/jupyter-deploy/jupyter_deploy/cli/host_app.py
```diff
@@ -218,6 +218,10 @@ def list_hosts(
         Path | None,
         typer.Option("--path", "-p", help="Directory of the project whose hosts to list."),
     ] = None,
+    role: Annotated[
```
JGuinegagne: I would say this is too template specific. How about `--pool-name str` ?

In this case, we need a generic, manifest-governed way of converting a `--pool-name` to a query param for the underlying API.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,135 @@
...
+            if not isinstance(item, dict):
+                continue
+            name = item.get("metadata", {}).get("name", "unknown")
+            limits = item.get("spec", {}).get("limits", {})
+            cpu_limit = limits.get("cpu", "-")
+            mem_limit = limits.get("memory", "-")
+            conditions = item.get("status", {}).get("conditions", [])
+            ready = next((c.get("status", "False") for c in conditions if c.get("type") == "Ready"), "Unknown")
+            node_count = item.get("status", {}).get("nodeCount", 0)
+            console.print(f"  [bold cyan]{name:<13}[/] {node_count:<8} {cpu_limit:<14} {mem_limit:<16} {ready}")
```
JGuinegagne: Let's discuss, I'm worried about being too template-specific.

Also, we should maintain consistency w/ other `list` cmds. IIRC `jd host list`, `jd server list` and `jd component list` don't return detailed information like this.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@pool_app.command(name="list")
+def list_pools(
+    project_dir: Annotated[
+        Path | None,
+        typer.Option("--path", "-p", help="Directory of the project."),
+    ] = None,
+    json_output: Annotated[bool, typer.Option("--json", help="Output as JSON.")] = False,
+) -> None:
```
JGuinegagne, line 26: sanity-check: do we need to paginate? I's most likely an overkill for pools, but just for visibility `list host` and `list server` commands handle pagination.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@pool_app.command()
+def status(
+    name: Annotated[str, typer.Option("--name", help="Name of the node pool.")],
+    project_dir: Annotated[
+        Path | None,
+        typer.Option("--path", "-p", help="Directory of the project."),
+    ] = None,
+    json_output: Annotated[bool, typer.Option("--json", help="Output as JSON.")] = False,
```
JGuinegagne, line 58: consistency again please: the other `status` cmds don't have `--json` flag.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,135 @@
...
+        console.print(f"[bold]Recent scaling events ({len(items)} nodes):[/]")
+        console.line()
+        for item in items:
+            if not isinstance(item, dict):
+                continue
+            metadata = item.get("metadata", {})
+            name = metadata.get("name", "unknown")
+            creation_time = metadata.get("creationTimestamp", "")
+            labels = metadata.get("labels", {})
+            pool = labels.get("karpenter.sh/nodepool", "unknown")
```
JGuinegagne: oof, we definitely can't hardcode this in the CLI

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if json_output:
+            console.print(json.dumps(details), highlight=False, markup=False, soft_wrap=True)
+            return
+
+        console.print_json(json.dumps(details))
+
+
+@pool_app.command()
+def scaling(
```
JGuinegagne: The "scaling" noun is a bit ambiguous.
- does it refer to `jd pool scale` > ie mutate the pool size
- or is it `jd pool scaling-events` > in that case, how about just `jd pool events` ?

Also, the 2nd one definitely requires some form of pagination or filter.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/cluster_handler.py
```diff
@@ -147,3 +151,43 @@ def show_cluster(self) -> ClusterDetail:
             endpoint=results.get("endpoint", ""),
             version=results.get("version", ""),
         )
+
+    def list_nodepools(self) -> list[Any]:
```
JGuinegagne: move to its own `PoolHandler` please

libs/jupyter-deploy/jupyter_deploy/handlers/resource/host_handler.py
```diff
@@ -38,17 +38,23 @@ def __init__(self, display_manager: DisplayManager) -> None:
...
 
     def list_hosts(
-        self, query: str, limit: int | None = None, continue_from: str | None = None
+        self, query: str, limit: int | None = None, continue_from: str | None = None, role: str | None = None
     ) -> tuple[list[str], str | None]:
         """Returns a list of host names and an optional continuation token."""
+        # Merge --role and --query into a single label selector passed to the manifest.
+        # --role is a convenience shorthand that expands to jupyter-deploy/role=<value>.
+        # If both are supplied they are joined with a comma (K8s AND semantics).
+        role_selector = f"jupyter-deploy/role={role}" if role else ""
```
JGuinegagne: nope, waaay too template-specific. We need a mechanism where the template manifest declaration drives the `--role` or `--pool-name` (my pref) conversion to the underlying query engine.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/helm.tf
```diff
@@ -204,5 +238,5 @@ resource "helm_release" "workspace_router" {
     },
   ]
 
-  depends_on = [module.node_group, helm_release.jupyter_k8s, null_resource.platform, helm_release.traefik_crds, null_resource.wait_for_lb_cleanup, kubernetes_namespace_v1.shared]
+  depends_on = [aws_eks_node_group.platform, module.karpenter, helm_release.jupyter_k8s, null_resource.platform, helm_release.traefik_crds, null_resource.wait_for_lb_cleanup, kubernetes_namespace_v1.shared]
```
JGuinegagne: this should just depend_on `null_resource.platform`, karpenter should be a member of that null resource

JGuinegagne (changes requested):
- look into cpu / memory limits for c, m and r instances, this doesn't look right
- for consistency, please use `platform_*` for helm charts / non-aws resources instead of modules
- we need to align on the `pool` and `host` commands, and their attributes. Mostly a/ keep consistency w/ existing commands, b/ watch out for commands that require pagination, c/ avoid leaking template-specific implementation to the CLI and handler layers

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/values.yaml
bhanderig: Good catch on the wording — `workspace_max_memory: 2048Gi` is a **NodePool fleet total**, not a per-node limit. Karpenter enforces it as a ceiling across all running workspace nodes combined, not on any individual instance. So `r` instances are fine here: a single `r6i.32xlarge` is 1TB RAM, two of those would hit the 2TB fleet cap. Updated the variable description in `variables.tf` to make this explicit.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
bhanderig: Removed the custom SG entirely — no separate SG is needed. The EKS cluster SG (created automatically by EKS) already allows all required node-to-cluster and inter-node traffic. `platform_karpenter.tf` now tags `module.eks_cluster.cluster_security_group_id` with `karpenter.sh/discovery` so the EC2NodeClass picks it up via selector. Matches the inference-clusters pattern.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/host_handler.py
bhanderig: Dropped both `--role` and `--pool-name`. The handler now takes `--query` as a free-form label selector passed directly to the manifest's `k8s.core.list-nodes` API call — no label expansion in the CLI or handler layer. Template-specific label keys (`jupyter-deploy/role`) live only in the manifest declaration where they belong.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
bhanderig: Removed entirely. `jd pool events` now calls the manifest's `pool.events` command which uses `k8s.custom.list-cluster` on NodeClaims — no label selectors or Karpenter-specific knowledge in the CLI or handler. The manifest owns the template-specific query.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/host_handler.py
JGuinegagne: `--pool-name` might be a valuable addition, but I agree to keep it to a future PR

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
JGuinegagne: Let's drop `pool events` from this PR — it currently just lists live NodeClaims rather than actual provisioning/consolidation events, so a real events view (with the pagination/filtering you flagged) is better as a dedicated follow-up.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
JGuinegagne: Let's insist on the `list[dict[str,str]]` shape now so we can add GPU/multi-category workspace pools without a breaking variable rename later — note this also means the chart's `nodepools.yaml`/`ec2nodeclasses.yaml` need to `range` over the list (routing can stay a singleton scalar; it's just the workspace pools that become the list).

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
JGuinegagne: Agreed this is high given user data lives on the separately-mounted EBS volume — let's reduce the root volume to something closer to the base template's default.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
bhanderig: Done — implemented as `list(map(string))` which the jd tfvars serializer already supports (same type used for EBS/EFS mounts in the base template). Each entry has `name`, `instance_families` (comma-separated string e.g. `c6i,m6i,r6i,c7i,m7i,r7i`), `disk_size_gb` (numeric string, validated with `can(tonumber())`), `max_cpu`, `max_memory`. Adding a GPU pool = one new list entry, no new variables.

AGENT.md
```diff
@@ -80,6 +80,23 @@ Code: `./libs/jupyter-deploy-tf-aws-eks-oidc`
 - cloud provider: `aws`
 - identity provider: `github` (via Dex OIDC)
 
+**IMPORTANT: When testing local chart changes against a deployed cluster, run `make deploy-aws-oidc`
```
andrii-i: This PR pins `workspace_router_chart_version = "0.1.1"`, but the published 0.1.1 chart predates the KEDA work, so a fresh `jd up` from this branch installs the routing tier with no ScaledObjects and traefik at `replicas: 1` — the autoscaling this PR adds silently does not deploy (rendering the published 0.1.1 chart with this PR's set flags produces no ScaledObjects). We should merge jupyter-k8s-aws#71, cut a chart release from it, and bump the pin here before merging, then this workaround paragraph can go.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -0,0 +1,103 @@
...
+          operator: In
+          values: ["on-demand"]
+        - key: kubernetes.io/arch
+          operator: In
+          values: ["amd64"]
+  limits:
+    cpu: {{ .Values.workspaceLimitsCpu | quote }}
+    memory: {{ .Values.workspaceLimitsMemory | quote }}
+  disruption:
+    consolidationPolicy: WhenEmptyOrUnderutilized
```
andrii-i: A running notebook session can be killed by node churn here, and workspace pods have no protection (no PDBs, no `karpenter.sh/do-not-disrupt`). Three sources, each with a fix:
- `WhenEmptyOrUnderutilized` lets consolidation evict a pod to bin-pack a partially-used node → use `WhenEmpty` (idle shutdown already empties nodes, so scale-to-zero still works).
- `expireAfter: 504h` force-drains at expiry (in v1.3 expiration drains immediately, pre-spins no replacement, ignores budgets and `do-not-disrupt`, [v1.3 docs](https://github.com/aws/karpenter-provider-aws/blob/v1.3.3/website/content/en/docs/concepts/disruption.md)) → set `expireAfter: Never`, or keep rotation with a `terminationGracePeriod`.
- `al2023@latest` drift-replaces nodes on every new AMI ([docs](https://github.com/aws/karpenter-provider-aws/blob/v1.3.3/website/content/en/docs/tasks/managing-amis.md) advise against `@latest` in prod) → pin the alias to a version.

The `node_expire_after` description in `variables.tf` then needs to match; it currently says rotation respects PDBs and waits for a replacement, which expiration in 1.3.3 does not do.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
```diff
@@ -0,0 +1,76 @@
...
+    - tags:
+        karpenter.sh/discovery: {{ .Values.clusterName | quote }}
+  securityGroupSelectorTerms:
+    - tags:
+        karpenter.sh/discovery: {{ .Values.clusterName | quote }}
+  tags:
+    karpenter.sh/cluster-name: {{ .Values.clusterName | quote }}
+  metadataOptions:
+    # hop 2: routing runs trusted infra pods only.
+    httpPutResponseHopLimit: 2
```
andrii-i: `httpPutResponseHopLimit: 2` lets any container on a routing node reach IMDS and read the node role's credentials, and no routing pod needs that: cert-manager and external-dns get AWS access via Pod Identity ([iam.tf](https://github.com/jupyter-infra/jupyter-deploy/blob/ed412e4e0695cb1823c832563cdf17f5164903cd/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf#L125-L140)), which does not use IMDS, and traefik/dex/oauth2-proxy/authmiddleware/web-app make no AWS calls. The comment's assumption that only our own pods run here is not enforced — `Workspace.spec` lets a user set `nodeSelector` and `tolerations` on their own workspace ([v0.2.0 types](https://github.com/jupyter-infra/jupyter-k8s/blob/v0.2.0/api/v1alpha1/workspace_types.go#L181-L187)), so a workspace user can schedule their notebook pod onto a routing node and read those node-role credentials (ECR pull, CNI, SSM). We should set `httpPutResponseHopLimit: 1` here, matching workspace-cpu; host-network agents (CNI, EBS CSI, Pod Identity agent) still reach IMDS at hop 1, so nothing on routing nodes loses access.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
```diff
@@ -32,25 +32,27 @@ workspace_app_jupyterlab_app_type        = "jupyterlab"
...
-    max_size      = "5"
-    desired_size  = "2"
-  }
-]
+platform_instance_types = ["m5.large"]
+platform_disk_size_gb   = 50
+platform_min_size       = 2
+platform_max_size       = 3
+
+karpenter_version  = "1.3.3"
```
andrii-i: These pins are old: karpenter `1.3.3` is from 2025-03 (the 1.3 line is at `1.3.8`, latest `1.14.0`), keda `2.16.1` from 2024-12 (latest `2.20.1`), prometheus `27.0.0` (latest `29.x`), so the feature starts well behind on fixes. We should take the `1.3.8` patch now and open a tracking issue for the minor upgrades, since `modules/karpenter/iam.tf` has to be re-synced when jumping minors (upstream split the controller policy in v1.9). Nonblocking.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/local-destroy-cleanup.sh.tftpl
```diff
@@ -22,6 +27,77 @@ CLUSTER_NAME="${cluster_name}"
...
 SLEEP_INTERVAL=10
 
+# --- Phase 0a: Uninstall Karpenter Helm release while controller is still alive ---
+#
+# Helm uninstall must happen BEFORE Karpenter nodes are terminated. If nodes are
+# gone first, the Karpenter controller pod dies and Helm can't reach its webhook,
+# causing "failed to delete release: karpenter" during terraform destroy.
+
+echo "Uninstalling Karpenter Helm release..."
+aws eks update-kubeconfig --name "$CLUSTER_NAME" --region "$REGION" > /dev/null 2>&1 || true
```
andrii-i: Phase 0a runs `aws eks update-kubeconfig` without `--kubeconfig`, so `jd down` overwrites the user's default `~/.kube/config` and switches their current context — this is also why AGENT.md now tells users to run update-kubeconfig manually before `jd down`. We should use the same `mktemp` + `KUBECONFIG` pattern as the three scripts this PR already hardened.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -0,0 +1,147 @@
+import json
```
andrii-i, line 1: `pool_app.py` and `pool_handler.py` land without unit tests, while every sibling command group has them (`test_cluster_app.py`, `test_host_handler.py`, ...); the str-vs-parsed result handling in `list_pools`/`show_pool` is the kind of contract those suites pin. We should add `test_pool_app.py` and `test_pool_handler.py` following the existing patterns. Nonblocking.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_host.py
```diff
@@ -70,22 +70,95 @@ def test_host_list_paginate_full(e2e_deployment: EndToEndDeployment) -> None:
...
     )
-    assert len(filtered["hosts"]) < all_count, (
-        f"Expected fewer hosts with filter, got {len(filtered['hosts'])} out of {all_count}"
+    assert len(filtered["hosts"]) < all_count, f"Expected query filter to return fewer than all {all_count} nodes"
+
+
+def test_host_list_query_routing_returns_only_routing_nodes(e2e_deployment: EndToEndDeployment) -> None:
+    """--query filters to Karpenter routing nodes using the node role label.
+
+    The routing NodePool always has at least 2 nodes (minReplicas=2 via KEDA).
```
andrii-i: The docstring says the routing pool always has at least 2 nodes via KEDA `minReplicas=2`, but that guarantees two traefik pods, not two nodes — anti-affinity is preferred-only, so consolidation can pack the tier onto one node and the `>= 2` assertion flakes. We should assert `>= 1` and drop the node-count claim from the docstring. Nonblocking.

.gitignore
```diff
@@ -175,3 +175,4 @@ test-results/
 # E2E test temporary files
 docker-compose.e2e-override.yml
 results*.log
+my-project-fresh-3/**/*
```
andrii-i: `my-project-fresh-3/**/*` is a local scratch-project name in the repo-root gitignore. We should drop it. Nonblocking.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_prometheus.tf
```diff
@@ -0,0 +1,87 @@
...
+      value = "false"
+    },
+    # Scrape interval — balance between freshness and cardinality
+    {
+      name  = "server.global.scrape_interval"
+      value = "30s"
+    },
+    # Retention — 15 days is sufficient for scaling signal history
+    {
+      name  = "server.retention"
```
andrii-i, line 55: 15d retention on a 20Gi PVC with no `server.retentionSize` set: a cardinality jump fills the volume and crashloops Prometheus, which stalls KEDA scaling. We should set `server.retentionSize` a bit under the PVC size (e.g. `18GB`). Nonblocking.

andrii-i (commented):
Thanks, this looks good. A few suggestions below, most nonblocking; the chart-pin sequencing is the one worth settling before merge. Also note the PR description drifted from the implementation (`--role` was dropped for `--query`, `pool scaling` became `status`/`events`, and the `--query` fix isn't in this diff); I don't see this as a big problem, just FYI.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
JGuinegagne: Thanks Andrii - could we look into this Gaurav? Reaching IMDS is generally a bad practice, although it depends what permissions the EC2 instance role has.

Re workspace scheduling on routing nodes:
1. We could tighten the nodeSelector via templates in the operator would be good idea; we can track as a separate issue
2. We may be able update the platform / routing node to repel all nodes w/ a certain label (which the operator adds). Future PRs though.

.gitignore
JGuinegagne: +1, we already gitignore `sandbox-*`

AGENT.md
bhanderig: This comment should not have made it out to PR. This was my internal testing note

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
bhanderig: Agree, i had it as 1 but then second set of structural changes brought this back. Will change to 1 for this PR

AGENT.md
bhanderig: Removed the internal testing note. The chart pin itself is blocked on #71 merging + a chart release — no code change possible here until then.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
bhanderig: Done — workspace pool switched to `WhenEmpty`. Validated on fresh cluster: `test_karpenter_workspace_provisioning_and_scale_to_zero` passes (node scales 0→1→0 within 4m33s). Routing pool stays `WhenEmptyOrUnderutilized` since it has no user workloads.

Re `expireAfter` and AMI pinning: keeping `expireAfter: 504h` for now since workspace pods are ephemeral (idle shutdown drains them well before 21 days), and `al2023@latest` drift is acceptable while we're on a single AZ sandbox — will address in the version-bump follow-up (#319) when we move to a production topology.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/ec2nodeclasses.yaml
bhanderig: Done — both pools set to `httpPutResponseHopLimit: 1`. Routing node comes up healthy (14/14 host tests pass). Pod Identity agent works at hop 1 as expected.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/presets/defaults-all.tfvars
bhanderig: Done — bumped to 1.3.8. Opened #319 to track minor upgrades (karpenter 1.3→1.14, keda 2.16→2.20, prometheus 27→29) since the IAM policy split in v1.9 requires `iam.tf` resync.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/local-destroy-cleanup.sh.tftpl
bhanderig: Done — destroy cleanup now uses `mktemp` + `KUBECONFIG` env var. Also fixed two related destroy-reliability bugs found during validation: added `aws_eks_access_policy_association` to `depends_on` on all platform helm releases (prevents RBAC-removed-before-uninstall race), and set `timeout = 900` on `karpenter_nodepools` (node draining can exceed the default 300s).

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
bhanderig: Agreed — will add `test_pool_app.py` and `test_pool_handler.py` as a follow-up. Not blocking merge.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_host.py
bhanderig: Done — assertion changed to `>= 1` and docstring updated. 14/14 host tests pass on fresh cluster.

.gitignore
bhanderig: Done — removed.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_prometheus.tf
bhanderig: Done — `server.retentionSize = 18GB` added. Whichever of time (15d) or size retention triggers first wins, so a cardinality spike can't fill the PVC.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -0,0 +1,103 @@
...
+  limits:
+    cpu: {{ .maxCpu | quote }}
+    memory: {{ .maxMemory | quote }}
+  disruption:
+    # WhenEmpty (not WhenEmptyOrUnderutilized): workspace pods carry no PDB and
+    # no karpenter.sh/do-not-disrupt annotation, so underutilized-consolidation
+    # would be free to evict a running notebook to bin-pack. Restricting to
+    # WhenEmpty means a node is only reclaimed once idle-shutdown has drained its
+    # last workspace pod — scale-to-zero still works, running notebooks are safe.
+    consolidationPolicy: WhenEmpty
```
andrii-i, line 99: `WhenEmpty` closes the consolidation vector, but two others reach the same outcome and neither is governed by `consolidationPolicy`: `expireAfter: 504h` force-drains at 21 days, and `al2023@latest` drifts the node whenever AWS publishes a new AL2023 AMI. With no PDB on workspace pods, both evict an active pod (kernel lost, pod reschedules on a fresh node), so "running notebooks are safe" here holds only for consolidation. The AMI drift recurs on every AMI publish. Suggest pinning the workspace AMI alias and rolling it deliberately, and making `expireAfter` a conscious choice (leave it, `Never`, or add a `terminationGracePeriod`). Nonblocking.

andrii-i (dismissed):
Thank you for addressing feedback @bhanderig, this looks good to me. One nonblocking comment to note, fine to address separately.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
bhanderig: Good callout — adding both to the version-bump tracking issue (#319). For this PR: idle shutdown (default 60min) empties workspace nodes well before the 504h expiry, so expireAfter is unlikely to hit a running session in practice. AMI drift is the more real risk for long-running notebooks; pinning the alias and rolling deliberately is the right answer there. Will address in the follow-up.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/values.yaml
```diff
@@ -1,11 +1,25 @@
...
 clusterName: ""
 nodeInstanceProfile: ""
+
+# Node expiry — forces periodic node rotation to prevent config drift.
+expireAfter: "504h"
+
+# ── Routing NodePool ──────────────────────────────────────────────────────────
+routing:
+  instanceCategories: ["c", "m"]
+  instanceGenerationMin: "5"
```
JGuinegagne, line 13: future PRs: would be better to use 6 here.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
JGuinegagne, line 6: sanity-check: still keeping a Karpenter module?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/helm.tf
```diff
@@ -35,7 +35,7 @@ resource "kubernetes_namespace_v1" "shared" {
     }
   }
 
-  depends_on = [aws_eks_access_policy_association.admin_role, aws_eks_access_policy_association.admin_user, aws_eks_node_group.platform, module.karpenter]
+  depends_on = [aws_eks_access_policy_association.admin_role, aws_eks_access_policy_association.admin_user, aws_eks_node_group.platform, helm_release.karpenter]
```
JGuinegagne, line 38: non-blocking: this is odd that this ns depends on `karpenter` release honestly

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/karpenter.tf
```diff
@@ -1,115 +1,21 @@
...
-  karpenter_version    = var.karpenter_version
   controller_role_arn  = module.karpenter_controller_role.role_arn
   node_role_arn        = module.karpenter_node_role.role_arn
-  vpc_id               = module.vpc.vpc_id
-  combined_tags        = local.combined_tags
   resource_name_prefix = local.resource_name_prefix
+  combined_tags        = local.combined_tags
 
   depends_on = [null_resource.core_node_addons, aws_eks_node_group.platform]
 }
```
JGuinegagne, line 17: Did you intend to keep this `karpenter.tf`?

Context: it's kind of an anti-pattern. We either use `platform_karpenter.tf`, or we call the modules in `main.tf`

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -0,0 +1,212 @@
...
+# EKS access entry for Karpenter-provisioned nodes.
+# Nodes provisioned by Karpenter use the karpenter_node_role. They need an EKS
+# access entry so the K8s API server trusts them and maps them to the
+# system:bootstrappers group for node registration.
+resource "aws_eks_access_entry" "karpenter_node" {
+  cluster_name  = module.eks_cluster.cluster_name
+  principal_arn = module.karpenter_node_role.role_arn
+  type          = "EC2_LINUX"
+  tags          = local.combined_tags
+}
```
JGuinegagne, line 212: question: why are these resources defined here?

Sounds like they should live in the `karpenter` module if you wanna keep it. In any case, watch out for delete flow, you likely need to keep these alive while the `karpenter` helm chart is up

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -90,46 +78,30 @@ def status(
...
+
+        with simple_display_manager.spinner(f"Checking status for pool {name}..."):
+            details = handler.show_pool(name=name)
+
+        resource = details.get("resource", {})
+        conditions = resource.get("status", {}).get("conditions", [])
+        ready = next(
+            (c.get("status", "Unknown") for c in conditions if c.get("type") == "Ready"),
+            "Unknown",
+        )
```
JGuinegagne, line 106: that logic should move to the handler.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
```diff
@@ -0,0 +1,68 @@
...
+        """Returns detailed info for a named node pool."""
+        command = self.project_manifest.get_command("pool.status")
+        runner = self._runner()
+        cli_paramdefs: dict[str, ResolvedCliParameter[Any]] = {
+            "name": StrResolvedCliParameter(parameter_name="name", value=name),
+        }
+        runner.run_command_sequence(command, cli_paramdefs=cli_paramdefs)
+        return collect_results(runner, command)
+
+    def list_events(self) -> list[Any]:
```
JGuinegagne, line 61: sanity-check: do you still need this? Otherwise I would remove it.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
```diff
@@ -0,0 +1,68 @@
...
+            raise NotImplementedError(f"PoolHandler not implemented for engine: {self.engine}")
+
+    def _runner(self) -> cmd_runner.ManifestCommandRunner:
+        return cmd_runner.ManifestCommandRunner(
+            display_manager=self.display_manager,
+            output_handler=self._output_handler,
+            variable_handler=self._variable_handler,
+        )
+
+    def list_pools(self) -> list[Any]:
```
JGuinegagne, line 42: could you define return types? see [payload](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy/jupyter_deploy/handlers/payloads.py)

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+        else:
+            raise NotImplementedError(f"PoolHandler not implemented for engine: {self.engine}")
+
+    def _runner(self) -> cmd_runner.ManifestCommandRunner:
+        return cmd_runner.ManifestCommandRunner(
+            display_manager=self.display_manager,
+            output_handler=self._output_handler,
+            variable_handler=self._variable_handler,
+        )
```
JGuinegagne, line 40: missing a `def get_status()`

JGuinegagne (approved):
Few comments - approving in case you want to iterate.

However let's please respect the cli / handler responsibility layers and tighten typing before the next release.

Also, watch out for `modules/karpenter` vs `engine/karpenter.tf` vs `engine/platform_karpenter.tf`. Having all 3 doesn't seem right, and the destroy flow looks brittle.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
andrii-i: Note this still stands for potential tracking and follow-up, nonblocking.

andrii-i (approved):
Thank you for the updates @bhanderig. This looks good to me.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/values.yaml
bhanderig: Agreed — will bump to 6 in the next PR (gen 5 instances are EOL for most families). Keeping this PR frozen at the validated state.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
bhanderig: Yes, intentionally — it holds the SQS interruption queue + 4 EventBridge rules (core AWS infra that must exist before the helm chart installs). Agree it's not a great fit for `modules/` since it's a singleton. Will inline into `platform_karpenter.tf` in a follow-up PR — it requires rewiring the Terraform state addresses and re-validating the destroy path.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/helm.tf
bhanderig: The dependency is there because the namespace is created by the karpenter helm release (`create_namespace = true`). If other resources reference that namespace before it exists, Terraform errors. Agree it looks odd — consolidating the module call into `platform_karpenter.tf` in the follow-up will make this read more naturally.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/karpenter.tf
bhanderig: It's a bridge between the core-infra module (SQS/EventBridge, must exist before kubeconfig) and the helm release in `platform_karpenter.tf` (needs working cluster). Agree it's an anti-pattern — will merge the module contents directly into `platform_karpenter.tf` and delete both `karpenter.tf` and `modules/karpenter/` in the consolidation follow-up.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
bhanderig: The SG tag, instance profile, and access entry are here because they depend on outputs from both the module (role ARN) and the EKS cluster (SG ID, cluster name) — putting them in the module would create a circular dependency. Agree it's messy; inlining the module contents here resolves this. Will address in the same consolidation PR. Destroy ordering is covered by the `depends_on` chain (validated with full jd up + jd down cycle today).

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
bhanderig: Agreed — will move in the pool handler cleanup follow-up (same PR as return types + `get_status()` + unit tests).

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
bhanderig: Yes still needed — it dispatches to the manifest's `pool.list` / `pool.status` commands. Will clean up with proper return types and add `get_status()` in the follow-up PR.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
bhanderig: Will do — adding typed payloads in the pool handler follow-up.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
bhanderig: Adding in the same follow-up PR.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/modules/karpenter/main.tf
bhanderig: Opened #320 to track the full consolidation (inline karpenter module, delete karpenter.tf, pool handler cleanup, gen-min bump). Will take it as the next PR.

Issue #305, opened 2026-07-09
Serve the roborev review workflow as a reusable workflow


andrii-i:
Follow-up to jupyter-infra/jupyter-k8s#400. Adopting the review currently means each consumer repo carries a full copy of the workflow (jupyter-infra/jupyter-k8s#410, #296, jupyter-infra/jupyter-k8s-ui#43), and every change is then duplicated across all copies; the container uid fix (jupyter-infra/jupyter-k8s#432) had to be ported to all three by hand.

Move the workflow body to a reusable workflow (`workflow_call`) in jupyter-deploy, next to the review image and IAM terraform it already provides. A consumer repo then adds a ~20 line shim (triggers, concurrency, permissions, `uses: jupyter-infra/jupyter-deploy/.github/workflows/roborev-review.yml@main`) instead of maintaining its own copy.

The run role trust in `tf-aws-iam-ci` can then pin the OIDC `job_workflow_ref` claim with an org-scoped `sub` instead of the per-repo allowlist, so onboarding needs no terraform apply. Image, model and region variables can become org-level Actions variables, leaving a new consumer with just the shim, the `review` environment, and any repo-specific overrides.

Also make the unconfigured state visible. Today unset `REVIEW_*` variables skip the job silently, indistinguishable from a trigger failure (see #304). A run with missing variables should state that the review is not configured, name the missing variables, and point at the setup steps instead of skipping silently.

Plan: add the reusable workflow to jupyter-deploy, convert jupyter-k8s first and validate on a real PR, extend the role trust, shrink jupyter-deploy and jupyter-k8s-ui to shims.

## Alternatives considered

Copy synchronization instead of a single source. Org starter-workflow templates or a scheduled sync bot that PRs workflow updates to consumers keep N copies with propagation lag, and the copies can still drift between syncs (the three current copies already differ in quoting). They also do nothing for trust: the run role would still be a per-repo allowlist.

A composite action wrapping the review steps. Composite actions run inside the caller's job, so the caller repo would still own the job, its permissions, and its environment, and the OIDC token would carry no `job_workflow_ref` claim; that claim exists only for reusable workflows, so the workflow-ref trust pin is not implementable this way.

[...]

PR #318, opened 2026-07-20
feat: serve the roborev review workflow as a reusable workflow

andrii-i:
Part of #305. Adopting the review currently means each consumer repo carries a full copy of `roborev-review.yml`, and every change is ported by hand to all copies (the container uid fix jupyter-infra/jupyter-k8s#432 had to land three times). This PR makes the jupyter-deploy copy the single source: it stays this repo's own review entrypoint and becomes callable by consumer repos, which shrink to a small shim. jupyter-infra/jupyter-k8s converts first in jupyter-infra/jupyter-k8s#441 to validate on a real PR, per the #305 plan.

Reusable workflow (`.github/workflows/roborev-review.yml`)
- Adds a `workflow_call` trigger next to the existing `pull_request_target` and `workflow_dispatch`, so one file serves both modes. A shim declares its own triggers, concurrency, and permissions (including `id-token: write`) and calls this workflow at `@main`; the called run executes in the caller's context: its events, its `GITHUB_TOKEN`, its `review` environment, and its `REVIEW_*` variables.
- New `resolve configuration` job resolves each value as `REVIEW_*` repository or organization variable, else default. Missing required values now produce a warning annotation and a step summary naming the variables and the setup steps, replacing the silent job skip (#304 made that state indistinguishable from a trigger failure).
- Two new optional knobs, `REVIEW_MIN_SEVERITY` (default `medium`) and `REVIEW_REASONING` (default `standard`), replace the hardcoded CLI flags with the same defaults, so behavior is unchanged where nothing new is set. As CLI flags they override the `.roborev.toml` equivalents in CI; repo review policy (agents, review types, per-repo `review_guidelines`) stays in `.roborev.toml`.
- The review job itself is unchanged: base checkout only, PR head fetched as git objects, credentials split between runner and container, stale-head guard, always-resolved commit status.

Workflow-ref trust (`libs/jupyter-infra-tf-aws-iam-ci`)
- The `iam_role` module accepts `oidc_job_workflow_refs`; when set, the trust policy also requires the OIDC `job_workflow_ref` claim to match one of the pinned workflows.
- New template variable `review_trust_workflow_refs`: when set, the run role trusts any repo in the org, but only from a job running the pinned reusable workflow, so onboarding a repo needs no terraform change. Default `[]` keeps the existing `review_repos` allowlist, so the deployed roles are untouched until the variable is set and re-applied.
- README and `.github/AGENT.md` document both trust modes; with only the allowlist, the `review` environment protection requirement stands as before.

Consumers referencing `@main` run whatever this repo's main serves: a bad change here propagates to every consumer at once, where per-repo copies isolated it. The trade and its alternatives (tag pinning, copy sync) are recorded in #305; it is bounded by this repo's own review and merge gate, a called workflow that can only comment and set statuses, and a run role that can only pull the image and invoke Bedrock. Caller-supplied configuration is validated in the config job and fails closed, and refs-trust mode stays opt-in with environment protection required in both modes.

### Rollout

1. Merge; this repo's own reviews continue in entrypoint mode with no configuration change. Then run a `workflow_dispatch` smoke (`smoke: true`) here to confirm the config job and the OIDC to Bedrock chain on the new file.
2. Merge jupyter-infra/jupyter-k8s#441 and validate the called path live: one real PR reviewed end to end through the shim, plus a smoke dispatch from jupyter-k8s. The shim works under the currently deployed allowlist trust, which does not inspect the workflow file, so no AWS change is needed to stay live.
3. Optionally set `review_trust_workflow_refs = ["jupyter-infra/jupyter-deploy/.github/workflows/roborev-review.yml@refs/heads/main"]` via `jd config` and `jd up` to switch the run role to workflow-ref trust, and consolidate `REVIEW_*` as org-level Actions variables.
4. Shrink jupyter-k8s-ui to a shim (tracked in #305).


andrii-i:
Addressed in 9f18fc4. The config job now validates every caller-supplied value and fails closed: allowlists for severity, reasoning, and region, ARN and private-ECR regexes for the role and image, and newline rejection, which also covers the GITHUB_OUTPUT note. No configuration value is interpolated into run blocks anymore; everything flows through env with quoted references. On the trust framing: agreed the docs downgrade was wrong. Environment protection is documented as required in both modes again, and the refs-mode docs now state plainly that it extends the caller set to every repo in the org and should be enabled only if every org repo admin may spend the run role's capabilities. The mode remains default-off. One scope note: the credentials reachable in the exploit path allow exactly ECR pull and Bedrock invocation for one hour, and the GITHUB_TOKEN involved is the calling repo's own, so the impact is unaudited model spend by an org repo admin rather than broader AWS or cross-repo access; the fixes close the injection vectors regardless.

Issue #320, opened 2026-07-22
refactor: consolidate Karpenter module + pool handler cleanup


bhanderig:
## Context

Follow-up from PR #316 review (JGuinegagne). The current structure works but has
structural debt flagged in review.

## Terraform consolidation

- Inline `modules/karpenter/` (SQS queue, 4 EventBridge rules, queue policy) directly
  into `platform_karpenter.tf`
- Delete `engine/karpenter.tf` (thin wrapper, anti-pattern)
- Delete `engine/modules/karpenter/` directory
- Move SG tag, instance profile, access entry into the same file
- Update all `depends_on` references from `module.karpenter.*` to direct resource refs
- Bump `routing_instance_generation_min` default from 5 to 6

Requires full `jd up` + `jd down` validation (resource address changes affect state).

## Pool handler cleanup

- Move display/formatting logic from `pool_app.py` to handler
- Add typed return payloads (see `handlers/payloads.py` pattern)
- Add `get_status()` method
- Add `test_pool_app.py` + `test_pool_handler.py` unit tests

## References

- JGuinegagne comments on #316 (Jul 22)
- andrii-i comment #6 on #316 (pool tests)

Issue #319, opened 2026-07-21
Bump Karpenter, KEDA, and Prometheus chart versions


bhanderig:
## Context

PR #316 ships with conservative chart pins that are behind on patches/minors:

| Component | Current | Latest (at time of writing) | Notes |
|-----------|---------|----------------------------|-------|
| Karpenter | 1.3.8 | 1.14.x | Minor jump requires resync of `modules/karpenter/iam.tf` — upstream split the controller policy in v1.9 |
| KEDA | 2.16.1 | 2.20.x | |
| Prometheus | 27.0.0 | 29.x | |

## Action items

1. **Karpenter 1.3 → latest**: Review IAM policy changes between 1.3 and target version. The controller policy was split in v1.9 — `iam.tf` must be updated to match.
2. **KEDA 2.16 → latest**: Check for breaking changes in values schema.
3. **Prometheus 27 → latest**: Check for breaking changes in values schema.

Each bump should be validated with the E2E suite (`test_autoscaling.py`, `test_host.py`) on a fresh cluster.

## References

- andrii-i's review comment on #316: "take 1.3.8 patch now and open a tracking issue for the minor upgrades"
- Template path: `libs/jupyter-deploy-tf-aws-eks-oidc/.../engine/presets/defaults-all.tfvars`

PR #322, opened 2026-07-22
refactor: consolidate Karpenter module + pool handler cleanup

### LINKED ISSUE #227: feat: pool commands (list, status) for EKS node groups
Add `jd pool list` and `jd pool status --name <nodegroup>` commands for EKS templates.

## Context

The EKS template uses role-based managed node groups (`components` and `workspaces`). Currently there's no CLI way to inspect node group health or scaling state without `aws eks describe-nodegroup`.

## Scope

| Command | API | Purpose |
|---------|-----|---------|
| `pool list` | `aws.eks.list-nodegroups` | List managed node groups with status |
| `pool status --name` | `aws.eks.describe-nodegroup` | Node group details: instance type, scaling config, health issues |

## Implementation notes

- Only available when the manifest declares `pool` (EKS templates only)
- New runner: `AwsEksNodeGroupRunner` in `provider/aws/`
- Manifest declares known pools so `list` can show descriptions alongside API data
- Follow the same pattern as `cluster` commands (PR #219)

bhanderig:
Follow-up to #316, addressing the structural feedback from @JGuinegagne and @andrii-i (tracked in #320).

## 1. Terraform consolidation

- **Inline `modules/karpenter/`** (SQS interruption queue, EventBridge rules, controller IAM policy) into `platform_karpenter.tf`, and **delete both the module and the `karpenter.tf` wrapper**. The Karpenter infra is a deployment singleton, not a reusable resource, so it belongs next to its helm release — matching the `platform_cluster_autoscaler.tf` precedent.
- Move the Karpenter controller IAM (policy doc + policy + attachment) into `iam.tf` alongside the controller/node roles (core-infra tier 1); the SQS queue policy stays with the queue in `platform_karpenter.tf`.
- Replace `module.karpenter.queue_name` with direct resource references and rewire `depends_on` to the specific inlined resources. Create/destroy ordering is now explicit.
- **YAGNI:** drop the spot-only interruption resources (spot-interruption + rebalance rules/targets) — every NodePool runs `capacity-type: on-demand`, so those events can never fire. Keep state-change + scheduled-change for on-demand termination/maintenance draining.
- Bump `routing_instance_generation_min` default `5` → `6`.
- Carry over the `aws:SourceArn` confused-deputy rationale as a comment on the SQS queue policy.

## 2. Pool handler cleanup

- `show_pool()` returns a typed `PoolDetail`; `get_status()` derives status via manifest `pool-status-rules` (mirrors `server_handler`, keeps template-specifics out of the handler). The handler is fully provider-agnostic.
- Pool status vocabulary expanded to `Ready`/`Creating`/`Degraded` off the Karpenter `Ready` condition (True/Unknown/False); falls back to `Unknown` (not `""`) when no rules are declared.
- `jd pool list` shows indented names + `{"pools": [...]}` JSON, matching `jd host list` / `jd server list` (names only). Help text and show/status wording made generic (no host/node/workspace/routing nouns).
- Drop the dead `list_events()` method.

## Convention alignment

- `data.aws_region.current.id` (not deprecated `.name`), matching the rest of the engine.
- Direct `aws_iam_instance_profile.karpenter_node.arn` reference (the string-constructed ARN was a module-boundary artifact).

## Testing

**Unit:** 2338 pass. `ruff` (check + format) + `mypy` clean. `pool.list`/`pool.status` pinned in `test_manifest_yaml.py`.

**Live (fresh EKS cluster):** `jd up` clean deploy + `jd down` clean teardown (validates the rewired destroy graph), then:

```
$ jd pool list
  routing
  workspace-cpu

$ jd pool list --json
{"pools": ["routing", "workspace-cpu"]}

$ jd pool status --name workspace-cpu
Pool status: Ready
```

`jd pool show --name <pool> --json` returns `{name, status, resource}` with the full NodePool object. Earlier cycle also ran E2E `test_host.py` (14/14) + `test_karpenter_workspace_provisioning_and_scale_to_zero`.

## Out of scope

- **MNG + Karpenter pool unification** — `jd pool` covering EKS Managed Node Groups alongside Karpenter pools is deferred to #324. Doing it provider-agnostically needs new manifest engine primitives (`flatten-list`/`coalesce`/`when`), so it's its own change rather than handler-side branching.
- Chart/operator version-bump pins — separate version-bump PR (#319).
- In-place template upgrade / `moved` blocks: not needed — template-upgrade isn't a supported flow and the Karpenter eks-oidc template isn't released yet (confirmed with @JGuinegagne).


michaelnchin (commented):
Reviewed for security, architecture, and state-migration surfaces. **Validated as safe:**

- IAM policy is a statement-by-statement faithful transcription of the deleted `modules/karpenter/iam.tf` (12 statements, identical conditions); the one semantic change — `AllowInstanceProfileGet` referencing `aws_iam_instance_profile.karpenter_node.arn` directly instead of a string-built ARN — is a strict improvement.
- SQS confused-deputy protection (`aws:SourceArn` scoped to the 4 EventBridge rule ARNs) survived the move.
- Destroy ordering: the rewired `depends_on` (policy attachment + queue policy) keeps the controller's IAM and queue access alive through helm uninstall, which is what finalizer cleanup needs; EventBridge rules going away earlier only stops new interruption notices mid-teardown.
- `platform_karpenter.tf` placement matches the `platform_cluster_autoscaler.tf` precedent and AGENT.md's tier-2 definition; `asdict` serialization in `pool_app.show` matches `host_app.py`.
- Gen `5→6` bump: `Gt` is exclusive so routing now requires gen 7+ — consistent with `variables.tf`'s "(exclusive)" wording and the #316 review ask.

**Findings** (3 inline, all nonblocking): missing `moved` blocks for the module→root state address changes, empty-string pool status when a manifest lacks `pool-status-rules`, and stale status vocabulary in the CLI test mocks.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -1,11 +1,338 @@
...
+  }
+
+  statement {
+    sid       = "AllowEKSClusterActions"
+    actions   = ["eks:DescribeCluster"]
+    resources = ["arn:${data.aws_partition.current.partition}:eks:${data.aws_region.current.id}:${data.aws_caller_identity.current.account_id}:cluster/${module.eks_cluster.cluster_name}"]
+  }
+}
+
+resource "aws_iam_policy" "karpenter_controller" {
```
michaelnchin: **[Reliability]** The inlining changes every Karpenter resource's state address (`module.karpenter.aws_sqs_queue.karpenter_interruption` → `aws_sqs_queue.karpenter_interruption`) with no `moved` blocks. Invisible on a fresh `jd up`, but an existing deployment that refreshes its template files and re-applies gets delete+recreate of the queue, the 4 EventBridge rules, this policy, and the attachment. Since the policy name (`${prefix}-karpenter-controller`) and queue name (`${cluster}-karpenter`) are unchanged and the creates aren't ordered after the deletes, the apply can fail with `EntityAlreadyExists` — and mid-apply the running controller has no policy attached, so it can neither provision nor terminate nodes until a re-apply succeeds. If in-place template refresh isn't a supported flow, fine as-is; otherwise ~7 `moved` blocks turn the whole migration into a state no-op. Nonblocking.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
```diff
@@ -48,21 +49,25 @@ def list_pools(self) -> list[Any]:
...
         runner = self._runner()
         cli_paramdefs: dict[str, ResolvedCliParameter[Any]] = {
             "name": StrResolvedCliParameter(parameter_name="name", value=name),
         }
         runner.run_command_sequence(command, cli_paramdefs=cli_paramdefs)
-        return collect_results(runner, command)
+        results = collect_results(runner, command)
+        resource = results.get("resource", {})
+        rules = self.project_manifest.pool_status_rules
+        status = evaluate_status_rules(json.dumps(resource), rules) if rules else ""
```
michaelnchin: **[Observability]** Templates without a `pool-status-rules` block get `status=""` here, so `jd pool status` prints `Pool status: ` with nothing after it and `jd pool show --json` emits `"status": ""`. `server_handler.get_status` falls back to the raw command result when rules are missing; there's no equivalent raw string for pools, so suggest `else "Unknown"` — matching `evaluate_status_rules`' no-match sentinel — rather than a silent empty string that reads as a bug. Nonblocking.

libs/jupyter-deploy/tests/unit/cli/test_pool_app.py
```diff
@@ -0,0 +1,123 @@
...
+class TestPoolApp(unittest.TestCase):
+    runner = CliRunner()
+
+    def get_mock_pool_handler(self) -> tuple[Mock, dict[str, Mock]]:
+        mock_handler = Mock()
+        mock_list_pools = Mock(return_value=["routing", "workspace-cpu"])
+        mock_show_pool = Mock(
+            return_value=PoolDetail(
+                name="workspace-cpu",
+                status="True",
```
michaelnchin: **[Review-risk]** These mocks use the pre-rules status vocabulary: with `pool-status-rules` the handler now emits `"Ready"/"NotReady"/"Unknown"` (as `test_pool_handler.py` pins), but the CLI mocks return `"True"`. Tests pass either way since the CLI is a passthrough — suggest `status="Ready"` / `return_value="Ready"` so the mocked contract matches what the handler actually produces. Nonblocking.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -1,11 +1,338 @@
...
-# modules/karpenter/; this file owns only the Helm release that deploys the controller.
-#
-# The karpenter-nodepools local chart (EC2NodeClass + NodePool CRDs) is also here
-# because it depends on the Karpenter CRDs installed by the controller Helm release.
+# All Karpenter infrastructure in one file: IAM policy, SQS interruption queue,
+# EventBridge rules, controller Helm chart, and the NodePool/EC2NodeClass chart.
+# Follows the platform_*.tf convention — singleton resources that deploy helm
+# charts and their supporting AWS infra onto the cluster.
+
+# ── Karpenter controller IAM policy ──────────────────────────────────────────
```
JGuinegagne: Non-blocking: inference clusters [precedent](https://github.com/jupyter-infra/inference-clusters/blob/main/libs/inference-tf-aws-eks-karpenter/inference_tf_aws_eks_karpenter/template/engine/iam.tf#L83), I would prefer to move IAM resources to `modules/iam.tf`

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -1,11 +1,338 @@
...
+    detail-type = ["EC2 Spot Instance Interruption Warning"]
+  })
+  tags = local.combined_tags
+}
+
+resource "aws_cloudwatch_event_target" "karpenter_spot_interruption" {
+  rule      = aws_cloudwatch_event_rule.karpenter_spot_interruption.name
+  target_id = "KarpenterInterruptionQueueTarget"
+  arn       = aws_sqs_queue.karpenter_interruption.arn
+}
```
JGuinegagne: sanity-check: do we need the spot interruption related resource?

If not then I would prefer to use YAGNI here, any resource we add increase the risk to hit resource limits.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/manifest.yaml
```diff
@@ -68,6 +68,15 @@ server-status-rules:
   all:
   - path: .status.conditions[type=Available].status
     equals: "True"
+pool-status-rules:
+- display: Ready
+  all:
+  - path: .status.conditions[type=Ready].status
+    equals: "True"
+- display: NotReady
```
JGuinegagne: This is fine, although not very informative!

We could expand to the following:
- track Karpenter [other conditions](https://karpenter.sh/docs/concepts/nodepools/#statusconditions)
- use MNG [status](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-status)

JGuinegagne: Let's discuss, but I would vote for `Ready`, `Creating`, `Degraded`

libs/jupyter-deploy/tests/unit/cli/test_pool_app.py
```diff
@@ -0,0 +1,123 @@
...
+    @patch("jupyter_deploy.handlers.resource.pool_handler.PoolHandler")
+    @patch("jupyter_deploy.cmd_utils.project_dir")
+    def test_pool_show_passes_project_dir(self, mock_project_dir: Mock, mock_handler_class: Mock) -> None:
+        mock_handler, _ = self.get_mock_pool_handler()
+        mock_handler_class.return_value = mock_handler
+
+        result = self.runner.invoke(pool_app, ["show", "--name", "routing", "--path", "/tmp/my-project"])
+
+        self.assertEqual(result.exit_code, 0)
+        mock_project_dir.assert_called()
```
JGuinegagne: could you use `assert_called_with()` here?

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
JGuinegagne: While we are at it, the CLI uses `host`, not `node` for virtual machines.

For consistency, could we update the update all references, e.g. `list_pools` > `List host pools in the projects.`.

JGuinegagne (commented):
Potentially for a future PR, but ideally would fit in this one:
- `jd pool list` should show both the MNG _and_ the Karpenter node pools
- `jd pool show --name <>` and `jd pool status --name <>` should equally apply to Karpenter or MNG

Supporting this behavior may require evolution to the command parser (as in introduce conditional declarations in the manifest). I think it's quite important for a proper CX though.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -1,11 +1,338 @@
...
+    effect  = "Allow"
+    actions = ["sqs:SendMessage"]
+    principals {
+      type        = "Service"
+      identifiers = ["events.amazonaws.com", "sqs.amazonaws.com"]
+    }
+    resources = [aws_sqs_queue.karpenter_interruption.arn]
+    condition {
+      test     = "ArnEquals"
+      variable = "aws:SourceArn"
```
andrii-i, line 40: In the deleted module, this `aws:SourceArn` condition carried a comment explaining why it exists: the queue name is deterministic, so without the condition an EventBridge rule in any AWS account could target this queue and inject forged interruption events, and Karpenter would drain healthy nodes. The condition itself moved over unchanged, but the explanation did not. We should carry that comment over so the next reader knows the condition is a security boundary and not an accident, nonblocking.

andrii-i (dismissed):
Thanks @bhanderig, this looks good. Please find a couple of comments below worth noting, fine for a follow-up.

Small things living outside the diff. First, `engine/iam.tf:154` has a comment saying the controller policy is attached inside `modules/karpenter/iam.tf`; that module is deleted by this PR, so the comment should now point at `platform_karpenter.tf`. Second, on the deprecation cleanup mentioned in the description: `terraform validate` warns that `data.aws_region.current.id` is deprecated on the current provider as well, across the whole engine, with `.region` as the replacement. Staying consistent with the engine's existing style was the right call for this PR; moving the engine to `.region` can be its own cleanup later.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
bhanderig: chatted w/ Jonathan - we don't support in-place template upgrade, and the karpenter oidc template isn't released yet so there's no old `module.karpenter.*` state out there to migrate. only hits on a fresh deploy where it's a no-op. skipping moved blocks for now, can add them if template-upgrade ever becomes a thing.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
bhanderig: done, falls back to `Unknown` now instead of empty string so status never renders blank.

libs/jupyter-deploy/tests/unit/cli/test_pool_app.py
bhanderig: done, mocks return `Ready` now to match what the handler actually emits.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
bhanderig: done - moved the controller policy (doc + policy + attachment) into `iam.tf` next to the roles. kept the SQS queue policy with the queue since it references the eventbridge rule arns.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
bhanderig: yeah good call, everything's on-demand so spot-interruption + rebalance can never fire. dropped both rules/targets + their sourcearn entries, kept state-change and scheduled-change for on-demand drain/maintenance. left a note on how to add them back if we ever do spot.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/manifest.yaml
bhanderig: done - `Ready`/`Creating`/`Degraded` off the Ready condition now. also added managed-pool status rules mapping the MNG status enum to the same vocab (part of the MNG work below).

libs/jupyter-deploy/tests/unit/cli/test_pool_app.py
bhanderig: done, `assert_called_once_with(Path(...))` now.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
bhanderig: done, swapped node->host everywhere (help text, docstrings, --name help, spinner) + regenned the cli docs.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
bhanderig: done, added the comment back explaining it's a confused-deputy guard - queue name is deterministic so without the sourcearn condition any account's eventbridge rule could push fake interruption events and drain our nodes.

bhanderig:
couple things from the review bodies (not inline):

@michaelnchin - thanks for the security/state-migration pass, all 3 inline findings addressed.

@JGuinegagne - went ahead and did the unified `jd pool` in this PR. `pool list` shows both MNG + karpenter with a type column, and show/status work on either. didn't add conditionals to the manifest parser - kept the abstraction in the handler instead (manifest declares the per-infra commands, handler combines + routes by type, provider does the calls). tested on a fresh cluster, the platform MNG shows up next to the karpenter pools and status/show route right + normalize to the same Ready/Creating/Degraded vocab.

@andrii-i - fixed the iam.tf:154 comment, though bit differently than you said - ended up moving the policy *into* iam.tf (per Jonathan's tier-1 comment) so the comment now just says it's in the same file. agree on leaving the `.region` cleanup for its own PR.

libs/jupyter-deploy/jupyter_deploy/exceptions.py
```diff
@@ -403,6 +403,20 @@ def __init__(self, image_name: str, tag: str) -> None:
...
+    """Raised when a pool name matches neither a Karpenter nor a managed pool.
+
+    Attributes:
+        pool_name: The name that was looked up
+        valid_pools: List of valid pool names
+    """
+
+    def __init__(self, pool_name: str, valid_pools: list[str]) -> None:
+        self.pool_name = pool_name
+        self.valid_pools = valid_pools
```
andrii-i: `valid_pools` is captured here but never surfaces: `error_decorator.py` has no `PoolNotFoundError` block, so it falls to the generic `JupyterDeployError` catch and `jd pool status --name <typo>` prints only `Pool 'typo' not found.`. With MNG names now being long generated strings like `jupyter-deploy-eks-6c197255-platform`, a mistyped --name is a likely path. `ComponentNotFoundError` and `ImageNotFoundError` carry the same attribute and each has a handler printing the valid names plus a list hint:

https://github.com/jupyter-infra/jupyter-deploy/blob/9f8655b738f1287ccd20db2da80cbddac709020c/libs/jupyter-deploy/jupyter_deploy/cli/error_decorator.py#L198-L203

We should add the matching handler so the failure tells the user what pools exist, or drop the attribute.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/manifest.yaml
```diff
@@ -194,6 +241,38 @@ commands:
   - result-name: pool.status.resource
     source: result
     source-key: '[0].Resource'
+# EKS Managed Node Group pools. The PoolHandler queries both these and the
+# Karpenter commands above, tags each result with its kind, and presents them
+# through the same `jd pool list/show/status` interface.
+- cmd: pool.list-managed
```
andrii-i: `pool.list-managed` and `pool.status-managed` are the only pool commands the template's manifest test does not pin: `test_cluster_autoscaling_commands_declared` / `test_cluster_autoscaling_commands_have_results` cover `pool.list` and `pool.status`, added with #316 for the same purpose:

https://github.com/jupyter-infra/jupyter-deploy/blob/9f8655b738f1287ccd20db2da80cbddac709020c/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_manifest_yaml.py#L284-L307

The pin matters more for these two than for the Karpenter ones: the handler resolves `pool.list-managed` with `get_command_or_none`, so a renamed or misspelled declaration raises nothing, the MNGs just disappear from `jd pool list`:

https://github.com/jupyter-infra/jupyter-deploy/blob/9f8655b738f1287ccd20db2da80cbddac709020c/libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py#L65-L77

We should add the two commands and their result names to those tests so the declaration cannot drift silently.

libs/jupyter-deploy/tests/unit/cli/test_pool_app.py
```diff
@@ -0,0 +1,144 @@
...
+        mock_handler, mock_fns = self.get_mock_pool_handler()
+        mock_handler_class.return_value = mock_handler
+
+        result = self.runner.invoke(pool_app, ["list"])
+
+        self.assertEqual(result.exit_code, 0)
+        self.assertIn("routing", result.output)
+        self.assertIn("workspace-cpu", result.output)
+        self.assertIn("platform", result.output)
+        # Option B: the type column is shown so users can tell what each pool is.
```
andrii-i: Is this still needed? Should we remove this?

andrii-i (commented):
Thanks @bhanderig for following up on the comments, found a couple of other things to note, please give it a look below. I think the stale `test_pool.py` E2E assertions would be nice to fix before this goes in.

docs/source/reference/resource/pool.md
```diff
@@ -1,6 +1,6 @@
 # `pool`
 
-Interact with the node pools managing workspace and routing nodes.
+Interact with the host pools managing workspace and routing hosts.
```
JGuinegagne: non-blocking: too specific, we shouldn't use `workspace` and `routing` nouns in the CLI description.

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -35,30 +35,37 @@ def list_pools(
...
             return
 
-        if names:
-            for name in names:
-                console.print(f"[bold cyan]{name}[/]")
+        if pools:
+            name_width = max(len(p.name) for p in pools)
+            console.print(f"[bold]{'NAME':<{name_width}}  TYPE[/]")
+            for pool in pools:
+                console.print(f"[bold cyan]{pool.name:<{name_width}}[/]  {pool.type}")
```
JGuinegagne: hum, so `list` returns essentially a table...

While that's useful information, it's a deviation from `jd server list`, `jd host list`. Do you think that's the right behavior?

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -34,30 +35,37 @@ def list_pools(
...
 
-        with simple_display_manager.spinner("Listing node pools..."):
-            names = handler.list_pools()
+        with simple_display_manager.spinner("Listing host pools..."):
+            pools = handler.list_pools()
 
         if json_output:
-            console.print(json.dumps(names), highlight=False, markup=False, soft_wrap=True)
+            console.print(
+                json.dumps([{"name": p.name, "type": p.type} for p in pools]),
```
andrii-i: Two eks-oidc E2E tests pin the old shape of this output (`libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_pool.py`, outside this diff, so flagging here). They only run against a deployed sandbox, not in PR CI, so nothing is red on this PR; the next unfiltered E2E run fails both. The fix:

```python
# test_pool_list_json_contains_nodepool_objects: pool list --json now emits {name, type} objects
names = {p["name"] for p in pools}
for name in EXPECTED_NODEPOOLS:
    assert name in names, f"Expected NodePool '{name}' in JSON output, got: {pools}"

# test_pool_status_returns_named_nodepool_details: pool status now prints Ready/Creating/Degraded
assert "Ready" in output, f"Expected NodePool '{nodepool_name}' to be Ready:\n{output}"
```

The list test could also assert a `managed` entry now that MNGs appear. Happy to push this if easier.

https://github.com/jupyter-infra/jupyter-deploy/blob/9f8655b738f1287ccd20db2da80cbddac709020c/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_pool.py#L44-L48
https://github.com/jupyter-infra/jupyter-deploy/blob/9f8655b738f1287ccd20db2da80cbddac709020c/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_pool.py#L63-L65

libs/jupyter-deploy/jupyter_deploy/handlers/resource/pool_handler.py
```diff
@@ -5,15 +5,28 @@
...
 from jupyter_deploy.handlers.base_project_handler import BaseProjectHandler
 from jupyter_deploy.handlers.payloads import PoolDetail
 from jupyter_deploy.handlers.resource.resource_utils import collect_results, evaluate_status_rules
 from jupyter_deploy.provider import manifest_command_runner as cmd_runner
 from jupyter_deploy.provider.resolved_clidefs import ResolvedCliParameter, StrResolvedCliParameter
 
+# Pool kinds surfaced through the unified `jd pool` interface. "karpenter" is a
+# Karpenter NodePool (k8s CRD); "managed" is an EKS Managed Node Group (AWS API).
+POOL_TYPE_KARPENTER = "karpenter"
+POOL_TYPE_MANAGED = "managed"
```
JGuinegagne: woaw woaw, that's leaking template-specific information into the handler, we absolutely cannot do that.

libs/jupyter-deploy/jupyter_deploy/exceptions.py
```diff
@@ -403,6 +403,20 @@ def __init__(self, image_name: str, tag: str) -> None:
         super().__init__(f"Tag '{tag}' not found for image '{image_name}'.")
 
 
+class PoolNotFoundError(JupyterDeployError, ValueError):
```
JGuinegagne: Use [this exception](https://github.com/jupyter-infra/jupyter-deploy/blob/aaa6c4bdb0ddcee8f0d82e1d4819b51d829cd147/libs/jupyter-deploy/jupyter_deploy/exceptions.py#L424) please

libs/jupyter-deploy/jupyter_deploy/manifest.py
```diff
@@ -366,6 +369,14 @@ def get_command(self, cmd_name: str) -> JupyterDeployCommandV1:
             raise CommandNotImplementedError(cmd_name)
         return command
 
+    def get_command_or_none(self, cmd_name: str) -> JupyterDeployCommandV1 | None:
```
JGuinegagne: I don't think we need this method, let's discuss offline

libs/jupyter-deploy/jupyter_deploy/cli/pool_app.py
```diff
@@ -12,7 +12,7 @@
 from jupyter_deploy.handlers.resource import pool_handler
 
 pool_app = typer.Typer(
-    help="Interact with the node pools managing workspace and routing nodes.",
+    help="Interact with the host pools managing workspace and routing hosts.",
```
JGuinegagne: `Interact with pools of hosts where apps and components run.`

JGuinegagne (changes requested):
Architecture: anti-pattern - holding PR.

Let's discuss offline.

JGuinegagne:
High level, I reckon we need this kind of declaration in the manifest:
```yaml
- cmd: pool.list
  sequence:
  - api-name: k8s.custom.list-cluster
    arguments:
    - api-attribute: group
      source: literal
      value: karpenter.sh
    - api-attribute: version
      source: literal
      value: v1
    - api-attribute: plural
      source: literal
      value: nodepools
  - api-name: aws.eks.list-nodegroups
    arguments:
    - api-attribute: cluster-name
      source: output
      value: cluster_name
  - api-name: flatten-list
    arguments:
      - api-attribute: list-1
        source: result
        source-key: '[0].Items'
      - api-attribute: list-2
        source: result
        source-key: '[1].Items'
  results:
  - result-name: pool.list
    source: result
    source-key: '[3].Items'
```

JGuinegagne:
For `status`, we need something like:
```yaml
- cmd: pool.status
  flags:
    name: is-karpenter
    conditions:
    - operator: in
       arguments:
       - api-attribute: value
         source: cli
         source-key: name
       - api-attribute: collection
         source: output
         source-key: platform_mng_name
  sequence: 
  - api-name: k8s.custom.get-cluster
     when:
       flag-name: is-karpenter
     arguments:
       - api-attribute: group
         source: literal
         value: karpenter.sh
       - api-attribute: version
         source: literal
         value: v1
      - api-attribute: plural
        source: literal
        value: nodepools
     - api-attribute: name
        source: cli
        source-key: name
  - api-name: aws.eks.describe-nodegroup
     when:
       not-flag-name: is-karpenter
     arguments:
     - api-attribute: cluster-name
        source: output
        source-key: cluster_name
     - api-attribute: nodegroup-name
        source: cli
        source-key: name
  - api-name: coalesce
     arguments:
     - api-attribute: value1
        source: result
        source-key: '[0]'
     - api-attribute: value2
        source: result
        source-key: '[1]'
  results:
  - result-name: pool.status.name
    source: result
    source-key: '[3].Name'
  - result-name: pool.status.resource
    source: result
    source-key: '[3].Resource'
```

bhanderig:
synced offline w/ Jonathan - pulling the MNG unification out of this PR and keeping `jd pool` karpenter-only for now. handler's back to provider-agnostic (no type logic, everything from the manifest), and `pool list` now matches `host`/`server list` - just names + `{"pools": [...]}`. filed #324 for the managed+karpenter aggregation, which needs the manifest primitives (flatten-list/coalesce/when) to do it without leaking types into the handler.

so this PR is now just the terraform consolidation + pool handler cleanup. @JGuinegagne mind another look when you get a chance?

JGuinegagne (approved):

andrii-i (approved):
Thanks @bhanderig, my comments are either addressed or moot and this looks good in total as well. 

Note that PR title and description don't match the reduced scope (not a blocker for me, up to you).

PR #326, opened 2026-07-30
fix: open hint, logs trimming

### LINKED ISSUE #323: `jd open`: do not show "Having trouble?" help when the command is not implemented
When a template does not implement `open`, `jd open` prints the generic troubleshooting help *before* the not-implemented error:

```
jd open

Having trouble?
🔍 verify that your host is running: jd host status

❌ Command 'open' is not implemented in this template.
```

The "Having trouble? / verify your host is running" hint is misleading — the host is fine; the command simply is not implemented for this template. The not-implemented check should short-circuit and print only the `❌` line, without the troubleshooting help.

### LINKED ISSUE #325: jd component logs stalls (100% CPU, no output) for pods with long history
# `jd component logs` stalls (100% CPU, no output) for pods with long history

## Symptom

`jd component logs --name oauth2-proxy` appears to hang: the spinner spins,
CPU pegs at ~98%, and nothing is ever printed. `jd component logs --name
oauth2-proxy -- --tail=5` returns instantly. So it is not the fetch — it is
the volume.

## Root cause

Two compounding issues, neither of which is kubectl:

1. **No default `--tail`.** `k8s_core_runner._deployment_logs`
   (`provider/k8s/k8s_core_runner.py`) builds `kubectl logs <pod> -n <ns>`
   with no bound and captures the **entire** container history into one
   string via `run_cmd_and_capture_output`. For a 13-day-old oauth2-proxy pod
   this is ~21k lines / ~2.75 MB.

2. **Rich re-parses the whole blob.** `component_app.logs` prints it with
   `console.print(log_output)` (`cli/component_app.py:160`), which applies
   markup parsing + syntax highlighting to the multi-MB string. This is
   superlinear — raw `kubectl logs` returns the same data in ~1.4s, while the
   full `jd` command burns 86s+ of CPU and never finishes (timed out at 90s
   having printed nothing).

## Repro

- `kubectl logs <oauth2-proxy-pod> -n jupyter-k8s-router | wc -lc`
  → 21192 lines / 2749966 bytes in ~1.4s.
- `jd component logs --name oauth2-proxy` → 90s, EXIT=124, zero output.
- `jd component logs --name oauth2-proxy -- --tail=5` → instant.

## Proposed fix

Pass a default tail when the caller supplies none:

- In `_deployment_logs`, if `extra` contains no `--tail`/`--since` bound, add a
  sane default (e.g. `--tail=1000`). A user-supplied `--tail` (or `--tail=-1`
  for all) overrides it, so power users keep full control.
- Consider printing with `console.print(log_output, markup=False,
  highlight=False)` (or a plain writer) so log content is never interpreted as
  Rich markup and rendering stays linear — a defense-in-depth even with a tail.

## Context

Observed 2026-07-28 on an EKS OIDC template deployment (`sandbox-kube1`),
oauth2-proxy pod up 13 days with 6 restarts.

JGuinegagne:
This PR fixes two small CLI issues: a/ `jd component logs` stalling on pods with long history, and b/ `jd open` printing misleading troubleshooting hints when a template does not implement the command.

Closes: #325
Closes: #323

### User experience
- `jd component logs --name <component>` returns promptly on long-lived pods
  - defaults to the last 1000 lines when no bound is given; a user-supplied `--tail`/`--since` (including `--tail=-1` for all) still takes over
- `jd open` on a template that does not implement `open` prints only the `❌ ... is not implemented` line — no preceding "Having trouble?" hint

### Implementation
1. bound log volume by default in `<jd>/provider/k8s/k8s_core_runner`
    1.1. prepend `--tail 1000` unless `extra` already carries a `--tail`/`--since`/`--since-time` bound
    1.2. without a bound, `kubectl logs` captures the entire container history into one string (multi-MB on old pods), which is what stalled the command
2. print logs without Rich markup in `<jd>/cli/component_app`: arbitrary log text is never re-parsed, keeping rendering linear as defense-in-depth
3. gate the `jd open` troubleshooting hints on `url is not None` in <jd>/cli/app `open`

### Testing
- ran the e2e `components` test against an `eks-oidc` template to ensure no regression

### Screenshots
**jd open - unimplemented**
Screenshot 2026-07-30 at 9 33 40 AM


michaelnchin (approved):
Looks good, thanks Jonathan!

PR #334, opened 2026-08-04
fix: autoscaler, fluentbit to depend on access entries

### LINKED ISSUE #333: eks-oidc: fluent-bit SA missing access-policy depends_on breaks destroy
Latent destroy-ordering hazard (not a deterministic failure): `kubernetes_service_account_v1.fluent_bit` in `platform_logging.tf` lacks a `depends_on` for the admin access-policy associations. Terraform is therefore *free* to delete the EKS access entry before the SA, which fails `Unauthorized`. Single-run destroys usually succeed (scheduler happens to order the SA first), so find-takedown has worked — but the ordering is not guaranteed, and concurrent/interrupted destroys expose it reliably.

Its sibling `kubernetes_namespace_v1.shared` (`helm.tf`) already guards against this — see the comment there: access-policy associations must outlive all K8s resources.

**File:** `libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_logging.tf:33`

**Current:** `depends_on = [null_resource.cluster_addons, aws_eks_node_group.platform]`

**Fix:** add `aws_eks_access_policy_association.admin_role` and `aws_eks_access_policy_association.admin_user`, matching the namespace resource.

JGuinegagne:
This PR fixes incorrectly configured dependency in the `eks-oidc` template resources: a/ fluent-bit and b/ cluster autoscaler _must_ depend on the eks access entry, otherwise on destroy the access entries _may_ get destroy first, preventing terraform to then authenticate to the cluster.

Closes: #333 

Also in this PR:
- template unit test to ensure k8s resource `depend_on` the eks access entry points

In practice, this rarely happens because other leaf resources hold the eks access entries, but it did in a rare race condition I observed locally.

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_cluster_autoscaler.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_cluster_autoscaler.tf
@@ -171,11 +171,15 @@ resource "helm_release" "cluster_autoscaler" {
     },
   ]
 
+  # Access-policy associations must outlive this release — without them the Helm/K8s
+  # provider loses authorization and destroy fails with "Unauthorized".
   depends_on = [
     null_resource.cluster_addons,
     aws_eks_node_group.platform,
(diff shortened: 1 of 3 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #338, opened 2026-08-07
fix: e2e web-app, karpenter tests

JGuinegagne:
This PR fixes two `eks-oidc` E2E tests that broke against a live cluster: a/ the web-app workspace tests failing on a stale `get_by_label("Name")` locator after a "Display Name" field was added to the create form, and b/ the Karpenter scale-to-zero test asserting on the global workspaces-node count, which never reaches zero when the cluster hosts other long-lived workspaces.

### Implementation
1. `<pytest-jupyter-deploy>/workspaces/web_app` `create_default_workspace`: match the name input via `get_by_role("textbox", name="Name", exact=True)`; the form's new "Display Name" field made a substring label match resolve to a non-input node
2. `<eks-oidc>/tests/e2e/test_autoscaling` scale-to-zero: baseline the workspaces-node set before create, then assert only the node(s) Karpenter *added* terminate — never the global count
3. `_workspaces_node_count()` → `_workspaces_nodes()` returning the node-name set; extract `_workspace_pod_node()` helper
4. skip when the pod co-locates onto a pre-existing workspaces node (Karpenter adds nothing, so scale-to-zero is unobservable); a fresh CI cluster always provisions a new node
5. pre-test cleanup now polls on the workspace pod being gone rather than the node count

### Testing
- e2e: reran the 5 previously-failing tests against a live `eks-oidc` deployment: `test_web_app` 12 passed / 1 skipped; Karpenter test correctly skipped on a tenant-shared cluster (verified via Karpenter logs that the test's own node was provisioned and terminated)

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
@@ -146,7 +146,9 @@ def _workspace_pod_node() -> str:
         "-l",
         f"workspace.jupyter.org/workspace-name={_SCALE_WORKSPACE}",
         "-o",
-        "jsonpath={.items[0].spec.nodeName}",
+        # Wildcard index (not [0]) so an empty item list yields '' instead of a
+        # jsonpath "array index out of bounds" error (exit 1) when no pod matches.
+        "jsonpath={.items[*].spec.nodeName}",
     )
 
 
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
@@ -121,7 +121,8 @@ def test_routing_deployments_have_no_hardcoded_replicas(e2e_deployment: EndToEnd
 # ── Karpenter workspace node provisioning ────────────────────────────────────
 
 
-def _workspaces_node_count() -> int:
+def _workspaces_nodes() -> set[str]:
+    """Return the set of node names currently labeled as workspaces-role nodes."""
     output = _kubectl(
         "get",
         "nodes",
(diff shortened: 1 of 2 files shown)


libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
```diff
@@ -121,74 +121,99 @@ def test_routing_deployments_have_no_hardcoded_replicas(e2e_deployment: EndToEnd
...
+    """Return the node hosting the _SCALE_WORKSPACE pod, or '' if none is scheduled yet."""
+    return _kubectl(
+        "get",
+        "pods",
+        "-n",
+        WORKSPACE_NAMESPACE,
+        "-l",
+        f"workspace.jupyter.org/workspace-name={_SCALE_WORKSPACE}",
+        "-o",
+        "jsonpath={.items[0].spec.nodeName}",
```
andrii-i: During every pre-test cleanup poll, with no pod matching, `[0]` makes `kubectl` error instead of printing nothing. To reproduce:

```console
$ kubectl get pods -l nomatch=1 -o 'jsonpath={.items[0].spec.nodeName}'
error: error executing jsonpath "{.items[0].spec.nodeName}": ... array index out of bounds: index 0, length 0
# exit 1; {.items[*].spec.nodeName} exits 0 with empty output
```

Error is currently swallowed by the `except`:
https://github.com/jupyter-infra/jupyter-deploy/blob/cb42d9a9972c42cbf0256e8e7f382b5198693309/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py#L180

Use wildcard so the helper returns `''` when no pod matches:
```suggestion
        "jsonpath={.items[*].spec.nodeName}",
```

andrii-i (dismissed):
Thank you @JGuinegagne. The locator fix and the rescoped scale-to-zero assertion both make sense. One comment below on the `_workspace_pod_node()` jsonpath.

andrii-i (approved):
Thank you @JGuinegagne.

PR #340, opened 2026-08-12
fix: set ebs as default storageClass

### LINKED ISSUE #321: No default StorageClass — no-template workspaces get stuck with an unbound PVC
### Problem
On the EKS OIDC template, creating a Workspace **without a WorkspaceTemplate** (bare/no-template, e.g. the UI's "No template" option) leaves the pod `Pending` forever. Its PVC never binds:

    FailedBinding: no persistent volumes available for this claim and no storage class is set

### Root cause
The cluster ships `ebs-sc` and `gp2`, but **neither is marked the default** StorageClass. Template-backed workspaces work because the template injects `primaryStorage.defaultStorageClassName: ebs-sc` into the PVC. A no-template workspace sends `storage: { size: 10Gi }` with no `storageClassName`, so with no cluster default the PVC stays unbound.

### Suggested fix
Mark `ebs-sc` as the default StorageClass so empty-storageClass PVCs bind:

    storageclass.kubernetes.io/is-default-class: "true"

(set on the `ebs-sc` StorageClass the template provisions).

### Environment
jupyter-deploy EKS OIDC template; jupyter-k8s operator; K8s (EKS).

JGuinegagne:
This PR fixes no-template Workspaces on the `eks-oidc` template hanging `Pending` forever on an unbound PVC. The cluster shipped `ebs-sc` and `gp2` but marked neither default, so a Workspace whose PVC has no `storageClassName` had no class to fall back on (`FailedBinding … no storage class is set`). Marking `ebs-sc` the default StorageClass makes such classless PVCs bind.

**Note:** Template-backed Workspaces set the class explicitly and are unaffected.

Closes: #321

Also in this PR:
- tighten eks get-token method to satify new `mypy` requirements

### User experience
- a no-template Workspace's PVC now inherits `ebs-sc` and binds, instead of the pod hanging `Pending` forever

### Implementation
1. `<eks-oidc>/engine/helm.tf`: set `storageClass.ebs.isDefault=true` to router chart: the aws-oidc chart stamps `storageclass.kubernetes.io/is-default-class: "true"` on `ebs-sc`

### Testing
- e2e: update `test_workspace.py` in `eks-oidc` template; asserts `ebs-sc` is the default StorageClass and a classless PVC inherits it
- manual: on a live `eks-oidc` cluster, reproduced `FailedBinding … no storage class is set`; after `jd up`, both a raw classless PVC and the operator-created PVC of a storage-less-template Workspace get `ebs-sc` stamped

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/helm.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/helm.tf
@@ -123,6 +123,14 @@ resource "helm_release" "workspace_router" {
         name  = "storageClass.ebs.create"
         value = "true"
       },
+      # Mark ebs-sc the cluster default StorageClass so PVCs with no explicit
+      # storageClassName bind; e.g. a no-template Workspace, which sends storage
+      # without a class (#321). Template-backed workspaces set the class
+      # explicitly and are unaffected.
+      {
+        name  = "storageClass.ebs.isDefault"
+        value = "true"
+      },
       {
         name  = "storageClass.efs.create"
         value = "false"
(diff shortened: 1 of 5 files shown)


michaelnchin (approved):
LGTM, thanks @JGuinegagne !

PR #341, opened 2026-08-17
feat: gpu workspace support for eks-oidc template

### LINKED ISSUE #336: eks-oidc: GPU workspace support
## Problem

A deployed eks-oidc cluster cannot run a GPU workspace. No nodepool provisions GPU instances, nothing installs the NVIDIA device plugin that advertises `nvidia.com/gpu` to the scheduler, and the shipped `jupyterlab` template declares no accelerator bounds, so the UI shows no GPU control. The stack above the template is ready end to end (jupyter-infra/jupyter-k8s-ui#56, jupyter-infra/jupyter-k8s-ui#59): any extended-resource key a WorkspaceTemplate declares in `resourceBounds.resources` gets a form control, and `defaultTolerations` plus `defaultNodeSelector` drive placement. The whole gap is in this template, in three pieces: capacity, device advertisement, and a template that requests the device.

### Admin

Opt in by setting `enable_gpu_pool = true` and running `jd config` then `jd up`; on an existing deployment, first add the new variables to the project's `variables.yaml` `overrides:` block, since existing projects do not pick up new template variables automatically. First use on an account without prior GPU usage also needs the service-quota raise noted below. The pool then appears in `jd pool list/show/status`, holding zero nodes and costing nothing until a GPU workspace runs; tuning (families, `max_gpus`, disk) goes through the `workspace-gpu` list entry.

### User

The create page gains a second card, `JupyterLab GPU`, showing the fixed shape it grants: one NVIDIA GPU plus the pinned vCPU and memory, with the controls rendered at those values and not adjustable. The first start provisions a node and pulls the image, minutes rather than seconds. When the pool is at `max_gpus` or AWS has no capacity in the allowed families, the workspace stays Pending until another GPU workspace stops or is deleted. After 30 idle minutes the workspace stops and its node is reclaimed about a minute later, ending the spend; restarting brings the environment back from its persistent volume on a freshly provisioned node.

## Proposed solution

Three parts mirroring the three gaps, plus one flag that gates them all.

### 1. GPU capacity: a `workspace_nodepools` entry

The list was designed to carry one (the variable docstring's example appends a `workspace-gpu` entry). A separate `gpu_nodepool` variable or a managed node group would fork pool handling into two code paths right after workspace capacity consolidated on Karpenter-only, and a fixed-AZ node group would reintroduce the #300 stranding class that Karpenter avoids by provisioning a node in the workspace volume's AZ on demand. `jd pool list/show/status` picks the new entry up with zero CLI changes.

Two per-pool values that `charts/karpenter-nodepools` hardcodes need to become list keys:

- The node label and taint. `nodepools.yaml:66` gives every pool the same `jupyter-deploy/role=workspaces` pair, so today a GPU template can neither target a GPU-only pool nor keep CPU workspaces off it. The new keys default to today's strings: a changed label or taint on a NodePool marks its nodes as drifted and Karpenter replaces them, restarting running workspaces, so with the flag off the rendered CPU NodePool must stay byte-identical. The toleration for the GPU taint stays declared in the template. EKS enables the `ExtendedResourceToleration` admission plugin, but it auto-tolerates only taints keyed by the requested resource name, and it does so for every `nvidia.com/gpu`-requesting pod in the cluster, so a `nvidia.com/gpu`-keyed taint could not scope the pool to workspaces; a template-declared toleration also keeps placement readable from the template alone.
- A `max_gpus` limit key (optional per entry) alongside `max_cpu`/`max_memory`, mapping to the NodePool `limits` entry for `nvidia.com/gpu`, which Karpenter accepts natively. Capping GPU spend through CPU limits cannot tell a GPU node from a CPU node across families (`g5.xlarge` has the same 4 vCPUs as `g4dn.xlarge` at nearly twice the price); the GPU count is the number an admin budgets in.

No AMI knob: the `al2023@latest` alias resolves GPU instance types to the NVIDIA AL2023 variant on its own (Karpenter does this variant auto-discovery since v1.0.3; the template pins 1.13.1), so a per-pool AMI setting would only duplicate a built-in mechanism as permanent config surface.

### 2. Device advertisement: the NVIDIA device plugin as a platform component

A new `engine/platform_nvidia_device_plugin.tf` modeled on `platform_keda.tf`, registered in the cluster-addons barrier and gated by `enable_gpu_pool`, so non-GPU clusters run no extra daemonset. Its scheduling needs three values, not one: a node selector on the GPU pool's label, a toleration for the GPU pool's taint (the chart's default tolerations cover only `CriticalAddonsOnly` and the `nvidia.com/gpu` key, and a values-supplied list replaces the defaults), and `nvidia.com/gpu.present: "true"` among the GPU pool's node labels, which the chart's default affinity requires on clusters without node-feature-discovery. Missing any of the three leaves the daemonset unscheduled, `nvidia.com/gpu` capacity never advertised, and every GPU workspace Pending with no error. One readout caveat: with the pool scaled to zero the daemonset reports `desiredNumberScheduled == 0`, which `jd health` currently shows as Degraded (#337); that fix or a `TROUBLESHOOT.md` note ships alongside.

The bare plugin rather than the GPU operator is deliberate: the NVIDIA AL2023 AMI ships the driver and container toolkit with the nvidia containerd runtime as the default, which makes the operator's driver management redundant here (AWS's guidance for these AMIs is to install the device plugin separately and disable operator-managed drivers), and its MIG and GPU-feature-discovery extras are not v1 needs. GPU feature discovery and the DCGM exporter are the pieces worth adopting next: the `gpu.product` node label would let the UI show the GPU model, and DCGM utilization data is what a later GPU-sharing decision needs.

### 3. A GPU WorkspaceTemplate

A second template rather than GPU bounds on `jupyterlab`: a template has one `defaultNodeSelector` and one toleration set, so a single template would steer either every workspace to the GPU pool or none, and every plain-notebook creator would meet a GPU control on the general-purpose card. `charts/workspace-defaults` renders exactly one hardcoded template today (`workspaces.tf:110` pins `workspaceTemplate.name` to `jupyterlab`, downstream wiring reads that single template by name at `workspaces.tf:228`, and the chart hardcodes bounds and placement instead of reading values); it needs to render a list, with per-template bounds, placement, image, and idle policy coming from values.

The GPU template declares:

- `nvidia.com/gpu` bounds `min: 1, max: 1`, and cpu/memory bounds pinned (`min == max`, values set from the allocatable observed on the hardware run's `g4dn.xlarge`). A GPU workspace owns its node: the chart restricts workspace pools to instance sizes of 2, 4, 8, 16 or 32 vCPUs (`nodepools.yaml:81`), every `g4dn`/`g5` size in that set carries exactly one GPU, and the pool taint keeps other workspaces off, so cpu/memory choice would change nothing for the user except which instance Karpenter buys. Pinning all three axes renders the card as one fixed shape (the form shows disabled controls at the pinned values), and an oversizing selection cannot force a larger instance. The alternative, a small adjustable range admitting the next size up (`g4dn.2xlarge`, 8 vCPU and 32 GiB), buys memory headroom at roughly double the node cost and restores the oversizing choice the pinning exists to remove; widening is additive later if real sessions outgrow the shape. Raising the GPU `max` later travels with a per-pool instance-size range in the same change.
- `min: 1` rather than `0`: a zero-GPU workspace from this card would keep the card's node selector, occupying a GPU node it does not use, holding `max_gpus` quota (NodePool limits count provisioned node capacity), and never reclaimed, since `WhenEmpty` consolidation ignores its non-empty node. Pausing GPU spend while keeping the environment is what stop does; the EBS volume persists across it. This depends on the simple-edit fix jupyter-infra/jupyter-k8s-ui#63, which ships in the same UI release this feature already waits for.
- `defaultResources` matching the pinned shape, so the form seed and the admission default agree.
- `defaultTolerations` for the GPU pool's taint and `defaultNodeSelector` for its label.
- An idle-shutdown default of 30 minutes (the `jupyterlab` template uses 60): an idle hour on the cheapest GPU node costs $0.53.
- The same image as `jupyterlab`. The container toolkit injects the driver stack, so `nvidia-smi` and the CUDA-bundling framework wheels work on the stock image; a CUDA-preloaded default image is a separate decision once usage shows which stack to bake.

The template declares plain `nvidia.com/gpu` only: MIG profile keys are a non-goal for v1 (they need MIG-capable instances and partitioning config), and DRA stays gated on operator support per the jupyter-infra/jupyter-k8s-ui#56 discussion. Named compute presets (a per-template dropdown of curated shapes) are a UI/CRD design of their own, tracked separately; whether presets live in the template or in a separate object templates reference is that issue's first question, and this card stays compatible either way.

### 4. One flag: `enable_gpu_pool`, default false

Following the `enable_component_logging` pattern. The flag alone yields the working stack: with it on, the template renders a built-in `workspace-gpu` pool entry unless `workspace_nodepools` already carries an entry by that name, which then takes precedence as the customization path; the device plugin and the GPU template render only with the flag. So no configuration can produce the GPU card without capacity behind it, and a hand-added GPU entry without the flag is an inert pool that provisions nothing. The alternative wiring, the flag plus a hand-written entry with a plan-time `check` enforcing the pairing (the `admin_role_names` precedent), keeps `workspace_nodepools` the single visible source of pool truth, at the cost of a two-step enablement whose mismatch is caught only at plan time; the built-in entry makes the mismatch unrepresentable and keeps enablement to one variable, and the pool stays observable through `jd pool list/show/status`, which read the cluster rather than the variable. Flag off, existing deployments see zero deltas: no pool, no daemonset, no template, and a byte-identical CPU NodePool (the drift constraint above).

It ships off rather than on because the GPU template card in the UI is a promise: clicking it should work and be sanctioned. With a default-on pool, that first click provisions an on-demand GPU node the admin never chose to offer.

The built-in entry, extending the docstring example with the new keys:

```
{ name = "workspace-gpu", instance_families = "g4dn,g5", disk_size_gb = "100", max_cpu = "64", max_memory = "256Gi", max_gpus = "4" }
```

`g4dn,g5` on-demand: Karpenter provisions the cheapest instance that fits, so one-GPU workspaces land on `g4dn` (T4, 16 GiB) whenever it has capacity, `g5` adds capacity fallback and 24 GiB of VRAM headroom at no default cost, and `max_gpus = "4"` caps the fleet at about $2 per hour worst case at `g4dn.xlarge` on-demand rates, a deliberately small default: raising it is a one-line override, while a generous one silently sanctions spend. Spot is substantially cheaper and Karpenter-native, but `platform_karpenter.tf:57` removed the spot-interruption SQS/EventBridge plumbing with an explicit note to re-add it alongside a spot NodePool, and `g4dn` spot in us-west-2 was reclaimed hourly and then unfulfillable during the 2026-08-04 testing, the wrong first experience for GPU workspaces. Spot returns as its own change together with the interruption plumbing.

One prerequisite lives outside the template: the `Running On-Demand G and VT instances` service quota defaults to 0 on accounts without prior GPU usage, and Karpenter then cannot launch any GPU instance (`VcpuLimitExceeded`). The flag's docs and `TROUBLESHOOT.md` point at the quota.

## Sequencing

This builds on the #320 consolidation (in flight on `refactor/karpenter-module-consolidation`): the GPU threading touches exactly the files that refactor reshapes, so it lands on top rather than being written twice against pre-consolidation code. The consolidation PR opens first.

The UI has its own chain, and it gates more than cosmetics. A deployed web-app that predates jupyter-infra/jupyter-k8s-ui#59 still renders a card for the new template but no GPU control, so workspaces created from it carry the template's placement without the device request, and Karpenter provisions GPU nodes for workspaces that never use the GPU. `workspace_router_chart_version` (`defaults-all.tfvars:15`, the aws-oidc chart bundling the UI) must therefore move to the chart release that ships UI v0.1.5 before or with the template change.

## Verification

The end state is verified on real hardware as of 2026-08-04 (`g4dn.xlarge`, one T4): template tolerations propagate, the pod schedules onto the tainted node, and `nvidia-smi` inside the workspace container sees the device (jupyter-infra/jupyter-k8s-ui#60).

For the PR: chart render and unit tests (the `variables.yaml` sync tests and the nodepool key assertions in `tests/unit/test_variables_yaml.py`, the expected-pool list in `tests/e2e/test_pool.py`, plus a render assertion that flag-off output for the existing pools is byte-identical to today's), then one integration verification on a real `jd` deploy: create from the GPU template in the UI, the pod lands on a Karpenter-provisioned GPU node, `nvidia-smi` in the container shows the device, delete scales the node back down. Whether GPU integration runs recur (push-to-main, release) is a separate decision.

andrii-i:
This PR enables the eks-oidc template to run GPU workspaces end to end. A `workspace_nodepools` entry can now declare an `accelerator` and reference template configs from the new `workspace_templates` variable; for these entries the platform installs the NVIDIA device plugin (the component that makes a node's GPUs schedulable to pods) and creates one workspace template per referenced config, each rendered as a create-page option by the web UI.

Deriving templates from pool entries keeps configuration in one place and prevents pool/template mismatches. Every GPU pool is fenced by its own role taint, so CPU workspaces never occupy GPU nodes. Workspace nodes are consolidated and their capacity reclaimed once every workspace on the node has stopped or idled out (Karpenter's `WhenEmpty` consolidation policy), so nodes with running notebooks are never evicted to reclaim capacity and user workflows are never interrupted.

Closes: #336

### User experience

Admins enable GPU workspaces in one of two ways, depending on how much they want to customize the pools:

1. Admins who just want to offer GPU workspaces and don't need custom settings such as their own instance families, `max_gpus` cap or idle shutdown timeout set `enable_default_gpu_pool: true` in the `overrides:` block of `variables.yaml` and run `jd config` and `jd up`. This adds a default GPU pool and workspace template: the `workspace-gpu` pool provisions on-demand instances from the `g4dn`, `g5`, `g6`, `g6e`, `g7` and `g7e` families, capped at 4 GPUs total, and the `jupyterlab-gpu` template offers one NVIDIA GPU per workspace with the deployment's default idle shutdown (60 minutes stock). A spec like this should be enough for the majority of single-GPU notebook work such as CUDA development, fine-tuning and inference on models that fit one GPU.
2. Admins who need a custom GPU setup, such as `p` instance pools, several GPU tiers with different idle rules, or their own capacity caps, write `workspace_nodepools` entries and `workspace_templates` configs directly in the same `overrides:` block. The pool entry declares the hardware side: instance families, disk size, capacity caps and the `accelerator`. The template config declares the workspace side: GPU count, cpu/memory, idle minutes, display name and description. The entry's `templates` key connects the two; one pool can offer several configs, and a config can be reused by pools that share a role. Full examples live in the [`workspace_nodepools`](https://github.com/andrii-i/jupyter-deploy/blob/f2a9be356bec2f7bf6b961b96f99dd743c9a7e50/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/variables.tf#L248) and [`workspace_templates`](https://github.com/andrii-i/jupyter-deploy/blob/f2a9be356bec2f7bf6b961b96f99dd743c9a7e50/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/variables.tf#L376) docstrings and the [autoscaling docs page](https://github.com/andrii-i/jupyter-deploy/blob/f2a9be356bec2f7bf6b961b96f99dd743c9a7e50/docs/source/templates/aws-eks-oidc-template/autoscaling.md).

End-users pick a workspace type from plainly-named cards on the create page, with no pool or GPU knowledge needed: in the web UI, rendered by [jupyter-k8s-ui](https://github.com/jupyter-infra/jupyter-k8s-ui) (a separate package, not changed by this PR), each workspace template appears as one card.

### Implementation

1. `engine/variables.tf`: four new optional keys on `workspace_nodepools` entries: `accelerator` (device stack the pool provides; "nvidia" is the only supported value), `max_gpus` (fleet GPU ceiling; absent means unbounded; requires `accelerator`), `role` (node label/taint value; defaults to "workspaces", or to the pool name when `accelerator` is set, so each GPU pool is fenced by construction), and `templates` (comma-separated config names the pool offers).
2. `engine/variables.tf`: new `workspace_templates` variable, a list of named configs with `gpus`/`cpu`/`memory`, `idle_minutes`, `display_name` and `description`. Both variables keep the template's existing `list(map(string))` shape, so the new keys are optional strings; moving to `list(object)` with typed fields stays open for a later iteration if the config needs more structure or stricter validation.
3. `engine/workspaces.tf`: for each (pool entry, referenced config) pair, terraform creates one `WorkspaceTemplate` next to the existing `jupyterlab` one; the entry supplies placement (nodeSelector + toleration from its role), the config supplies the rest. A config with `gpus` must also set cpu and memory, and its card offers exactly those values with nothing for the user to adjust: the node is provisioned for the workspace, so a different cpu/memory pick would only change which instance Karpenter buys. A config without cpu/memory uses the same default values as the standard `jupyterlab` template. Several pools can offer the same config when they share a role; pools with different roles need separate configs, since the rendered template pins one nodeSelector.
4. `engine/platform_karpenter.tf`: `enable_default_gpu_pool` (renamed from `enable_gpu_pool`) injects the default `workspace-gpu` entry and `jupyterlab-gpu` config; combining the flag with hand-written accelerator entries raises a plan-time error.
5. `engine/platform_nvidia_device_plugin.tf`: the NVIDIA device plugin as a platform Helm release, installed whenever an accelerator entry exists; its DaemonSet targets nodes with the `nvidia.com/gpu.present` label and tolerates every GPU pool's taint, so several GPU pools work.
6. Scale-to-zero lifecycle: a GPU pool holds zero nodes and costs nothing until a GPU workspace starts, so the first start takes a few minutes (Karpenter provisions a node, the image is pulled). After the idle timeout (the deployment's `workspaces_idle_shutdown_timeout_default`, 60 minutes stock, shared with the `jupyterlab` template) the operator stops the workspace: its pod is deleted and the emptied node is reclaimed, while the workspace's EBS volume is kept. Restarting the workspace provisions a fresh node and reattaches the volume, so users resume with all their files intact.
7. Admins never pick an AMI and no separate GPU image is maintained, which removes both the setup burden and the risk of driver/version mismatches: GPU nodes boot with working NVIDIA drivers on their own. The `al2023@latest` AMI alias resolves GPU instance types to the NVIDIA variant, the AMI carries the driver, the container toolkit injects it into workspace containers, and CUDA ships in framework wheels.
8. Misconfiguration fails at plan time, not on the cluster: the flag conflict, duplicate pool or config names, `templates` referencing a missing config, cross-role config reuse, a `gpus` config offered by a pool without an accelerator (its workspaces could never schedule), `idle_minutes` or the deployment idle default outside the idle-shutdown window, an accelerator pool sharing a role with a CPU pool or claiming the base `jupyterlab` template's `workspaces` role (either would let CPU pods onto GPU nodes), any pool claiming the routing pool's `routing` role, a pool set that leaves no pool on the `workspaces` role the base template schedules on, and the pairing rules (`max_gpus` requires `accelerator`, `gpus` requires cpu and memory).
9. `charts/karpenter-nodepools`: workspace pool entries gain optional `role`, `gpu` and `maxGpus` values (`role` sets the node label and taint value, `gpu: "true"` adds the `nvidia.com/gpu.present` node label, `maxGpus` renders a `nvidia.com/gpu` limit); terraform maps `accelerator: nvidia` onto them. Entries without the new keys render exactly the same objects as before, so nothing changes for existing deployments (verified by tests).
10. AWS prerequisite: accounts without prior GPU usage have the `Running On-Demand G and VT instances` quota at 0 and need it raised before the first GPU node can launch; documented in the variable docstring and TROUBLESHOOT.

### Testing

- Added unit tests covering the default GPU pool and template rendering, custom pool scenarios
- Added e2e tests, opt-in via `JD_E2E_GPU_ENABLED`, testing the GPU workspace lifecycle and kernel-level CUDA
- Verified the user flows on a real deployment with GPUs available: deploy with the default GPU pool (`enable_default_gpu_pool: true`), a custom GPU pool with its own template config, autoscaling from and back to zero. Also tested the web UI flows end to end: create/open/stop/restart of a GPU workspace with files surviving onto a fresh node, automatic idle shutdown. Note that creating a GPU workspace from the card needs jupyter-infra/jupyter-k8s-ui#69 fixed first (verified with the GPU request patched onto the workspace).

Next steps beyond this PR:
- jupyter-infra/jupyter-k8s-ui#69
- `mig_profile` on a pool entry to offer fractions of a GPU, a `namespaces` field to copy a template into more workspace namespaces, and per-config images for other IDEs and non-NVIDIA hardware.
- Optional-component support in the CLI: currently GPU components do not appear in `jd component`/`jd health` because the manifest has no way to mark a component optional

### Screenshots

Create page with the GPU card selected: both template cards, locked image, pinned resources, 30-minute idle default
Screenshot 2026-08-20 at 1 26 03 AM
Screenshot 2026-08-20 at 1 26 11 AM

Running GPU workspace card
Screenshot 2026-08-20 at 1 36 08 AM

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -184,8 +184,9 @@ jd server logs --name <workspace-name> --scope <namespace>
 ```
 
 A workspace stuck in `Starting` is often waiting on a node. GPU workspaces require
-`enable_gpu_pool: true` (the `workspace-gpu` Karpenter pool plus the NVIDIA device
-plugin) and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
+a GPU pool entry in `workspace_nodepools` (`enable_gpu_pool: true` provides the
+built-in one: the `workspace-gpu` Karpenter pool plus the NVIDIA device plugin)
+and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
 component reads `Degraded` in `jd health` while the GPU pool is scaled to zero:
 a DaemonSet with desired count 0 is idle, not failing.
 
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -184,9 +184,10 @@ jd server logs --name <workspace-name> --scope <namespace>
 ```
 
 A workspace stuck in `Starting` is often waiting on a node. GPU workspaces require
-a GPU pool entry in `workspace_nodepools` (`enable_default_gpu_pool: true` provides the
-built-in one: the `workspace-gpu` Karpenter pool plus the NVIDIA device plugin)
-and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
+an accelerator pool entry in `workspace_nodepools` with a template config offered
+through its `templates` key (`enable_default_gpu_pool: true` provides the built-in
+pair: the `workspace-gpu` Karpenter pool, the NVIDIA device plugin, and the
+`jupyterlab-gpu` card) and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
 component reads `Degraded` in `jd health` while the GPU pool is scaled to zero:
 a DaemonSet with desired count 0 is idle, not failing.
 
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -19,8 +19,9 @@ this template:
 
 - **EC2 vCPUs** — node group scaling fails with `VcpuLimitExceeded`. New accounts
   start with a low On-Demand vCPU limit per region, and `0` for accelerated
-  families (GPU/Neuron), which blocks `workspaces` node groups that use GPU
-  instance types. Quotas are measured in vCPUs, not instance counts.
+  families (GPU/Neuron), which blocks the workspace pools' GPU nodes
+  (Karpenter provisions one per GPU workspace). Quotas are measured in vCPUs,
+  not instance counts.
 - **Elastic IPs** — `AllocateAddress` fails with `AddressLimitExceeded`. The VPC
   provisions one NAT gateway (and therefore one EIP) per availability zone — two by
   default. The per-region limit defaults to 5.
@@ -40,9 +41,9 @@ Relevant quota codes:
 
 To resolve:
 
-- Pick smaller `instance_type` values in `node_groups`, reduce `desired_size`, or
-  free resources by destroying stale deployments (`jd down --path <project-dir>`),
-  then re-run `jd config` and `jd up`.
+- Pick smaller instance families or lower `max_cpu`/`max_gpus` in
+  `workspace_nodepools`, or free resources by destroying stale deployments
+  (`jd down --path <project-dir>`), then re-run `jd config` and `jd up`.
 - Or request a higher limit.
 
 {{ request-quota-increase-instructions }}
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -187,8 +187,9 @@ A workspace stuck in `Starting` is often waiting on a node. GPU workspaces requi
 an accelerator pool entry in `workspace_nodepools` with a template config offered
 through its `templates` key (`enable_default_gpu_pool: true` provides the built-in
 pair: the `workspace-gpu` Karpenter pool, the NVIDIA device plugin, and the
-`jupyterlab-gpu` card) and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
-component reads `Degraded` in `jd health` while the GPU pool is scaled to zero:
-a DaemonSet with desired count 0 is idle, not failing.
+`jupyterlab-gpu` card) and sufficient G/VT vCPU quota (see above). Inspect the
+device plugin with `kubectl get daemonset nvidia-device-plugin -n kube-system`;
+it reports desired count 0 while the GPU pool is scaled to zero, which is idle,
+not failing.
 
 {{ generated-by-footer }}
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -184,7 +184,7 @@ jd server logs --name <workspace-name> --scope <namespace>
 ```
 
 A workspace stuck in `Starting` is often waiting on a node. GPU workspaces require
-a GPU pool entry in `workspace_nodepools` (`enable_gpu_pool: true` provides the
+a GPU pool entry in `workspace_nodepools` (`enable_default_gpu_pool: true` provides the
 built-in one: the `workspace-gpu` Karpenter pool plus the NVIDIA device plugin)
 and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
 component reads `Degraded` in `jd health` while the GPU pool is scaled to zero:
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
@@ -183,8 +183,10 @@ jd server show --name <workspace-name> --scope <namespace>
 jd server logs --name <workspace-name> --scope <namespace>
 ```
 
-A workspace stuck in `Starting` is often waiting on a node. GPU workspaces require a
-`workspaces` node group with GPU capacity and sufficient vCPU quota (see above). If the
-node group `desired_size` is `0`, scale it up in `node_groups` and run `jd up`.
+A workspace stuck in `Starting` is often waiting on a node. GPU workspaces require
+`enable_gpu_pool: true` (the `workspace-gpu` Karpenter pool plus the NVIDIA device
+plugin) and sufficient G/VT vCPU quota (see above). The `nvidia-device-plugin`
+component reads `Degraded` in `jd health` while the GPU pool is scaled to zero:
+a DaemonSet with desired count 0 is idle, not failing.
 
 {{ generated-by-footer }}
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -217,7 +217,7 @@ resource "null_resource" "karpenter_nodepools_finalizer_cleanup" {
 }
 
 # ── GPU pool synthesis ────────────────────────────────────────────────────────
-# enable_gpu_pool alone yields working GPU capacity: when on and no entry named
+# enable_default_gpu_pool alone yields working GPU capacity: when on and no entry named
 # workspace-gpu exists, the built-in entry below is appended. A user-defined
 # workspace-gpu entry takes precedence as the customization path (issue #336).
 locals {
@@ -246,7 +246,7 @@ locals {
   }
   workspace_nodepools_effective = concat(
     var.workspace_nodepools,
-    var.enable_gpu_pool && !contains([for p in var.workspace_nodepools : p["name"]], local.gpu_pool_name)
+    var.enable_default_gpu_pool && !contains([for p in var.workspace_nodepools : p["name"]], local.gpu_pool_name)
     ? [local.gpu_pool_builtin] : [],
   )
 }
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -226,7 +226,7 @@ locals {
   gpu_pool_name = "workspace-gpu"
   gpu_pool_builtin = {
     name              = local.gpu_pool_name
-    instance_families = "g4dn,g5,g6,g6e"
+    instance_families = "g4dn,g5,g6,g6e,g7,g7e"
     disk_size_gb      = "100"
     max_cpu           = "64"
     max_memory        = "256Gi"
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -224,7 +224,7 @@ locals {
   gpu_pool_name = "workspace-gpu"
   gpu_pool_builtin = {
     name              = local.gpu_pool_name
-    instance_families = "g4dn,g5"
+    instance_families = "g4dn,g5,g6,g6e"
     disk_size_gb      = "100"
     max_cpu           = "64"
     max_memory        = "256Gi"
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -348,6 +348,15 @@ resource "helm_release" "karpenter_nodepools" {
       condition     = length(distinct([for p in local.workspace_nodepools_effective : p["name"]])) == length(local.workspace_nodepools_effective)
       error_message = "workspace_nodepools names must be unique, including the built-in \"workspace-gpu\" entry injected by enable_default_gpu_pool."
     }
+    precondition {
+      # Re-checks the variable-level role-collision rule over the effective
+      # list: the validation cannot see the entry enable_default_gpu_pool injects.
+      condition = length(setintersection(
+        toset([for p in local.workspace_nodepools_normalized : p["role"] if lookup(p, "accelerator", "") != ""]),
+        toset([for p in local.workspace_nodepools_normalized : lookup(p, "role", "workspaces") if lookup(p, "accelerator", "") == ""]),
+      )) == 0
+      error_message = "An accelerator pool must not share a role with a non-accelerator pool, including the built-in \"workspaces-gpu\" role injected by enable_default_gpu_pool: a shared role would let CPU workspaces onto GPU nodes."
+    }
   }
 
   depends_on = [
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -249,11 +249,6 @@ locals {
     var.enable_gpu_pool && !contains([for p in var.workspace_nodepools : p["name"]], local.gpu_pool_name)
     ? [local.gpu_pool_builtin] : [],
   )
-  # Role label of the GPU pool; the device plugin selects on it.
-  gpu_pool_role = try(
-    [for p in local.workspace_nodepools_effective : lookup(p, "role", "workspaces") if p["name"] == local.gpu_pool_name][0],
-    "workspaces-gpu",
-  )
 }
 
 resource "helm_release" "karpenter_nodepools" {
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -237,8 +237,9 @@ locals {
     accelerator = "nvidia"
     templates   = "jupyterlab-gpu"
   }
-  # cpu/memory target the g4dn.xlarge allocatable (an over-pin is permanently
-  # unschedulable; finalize against a live node, issue #336). Idle follows the
+  # cpu/memory sit under the g4dn.xlarge allocatable with headroom, so kubelet
+  # reservation or DaemonSet growth cannot leave the pod permanently
+  # unschedulable on the smallest family. Idle follows the
   # deployment default: a literal here could fall outside the admin's idle
   # window and fail every plan mentioning a config the admin never wrote.
   gpu_template_builtin = {
@@ -246,8 +247,8 @@ locals {
     display_name = "JupyterLab GPU"
     description  = "JupyterLab workspace with one NVIDIA GPU and persistent EBS storage"
     gpus         = "1"
-    cpu          = "3500m"
-    memory       = "13Gi"
+    cpu          = "3"
+    memory       = "12Gi"
     idle_minutes = tostring(var.workspaces_idle_shutdown_timeout_default)
   }
 
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
@@ -229,8 +229,10 @@ locals {
     max_cpu           = "64"
     max_memory        = "256Gi"
     max_gpus          = "4"
-    role              = "workspaces-gpu"
-    gpu               = "true"
+    # Explicit: the accelerator role default would yield "workspace-gpu" (the
+    # pool name), and tests, fixtures, and docs pin "workspaces-gpu".
+    role        = "workspaces-gpu"
+    accelerator = "nvidia"
     # Template keys: workspaces.tf derives the jupyterlab-gpu WorkspaceTemplate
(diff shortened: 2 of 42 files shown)


libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
```diff
@@ -0,0 +1,275 @@
...
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.rendered = _helm_template("charts/karpenter-nodepools", "karpenter-nodepools", "values_karpenter_cpu.yaml")
+
+    def test_render_matches_golden(self) -> None:
+        self.compare_to_golden(self.rendered, "golden_karpenter_nodepools_cpu.yaml")
+
+
+@_SKIP_WITHOUT_HELM
```
earaghbidikashani, line 92: non-blocking:
The GPU NodePool is subject to the same Karpenter spec.template hash-drift invariant this module's docstring cites for the CPU pool, but its GPU-specific bytes (the nvidia.com/gpu.present label block, the nvidia.com/gpu limit) are only field-asserted, not byte-pinned. 

Consider a `golden_karpenter_nodepools_cpu_gpu.yaml` byte-compared like the CPU golden, so a stray requantization or unquoted value on a GPU pool is caught.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -0,0 +1,177 @@
...
+            return
+        time.sleep(interval_s)
+    raise TimeoutError(f"Condition not met within {timeout_s}s: {msg}")
+
+
+def _gpu_node_count() -> int:
+    return len(get_node_names(GPU_ROLE_SELECTOR))
+
+
+@skip_if_testvars_not_set(["JD_E2E_GPU_ENABLED"])
```
earaghbidikashani, line 56: `JD_E2E_GPU_ENABLED` is set in no committed workflow that I see, and base.yaml defaults `enable_default_gpu_pool` to false, so these GPU e2e tests skip in every automated lane. 

Since they're also the only coverage of platform_nvidia_device_plugin.tf's tolerations/nodeSelector rendering, the GPU runtime path can regress silently. 
Is a scheduled/manual GPU lane planned that sets the gate + `JD_E2E_VAR_ENABLE_DEFAULT_GPU_POOL=true`?

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -216,6 +216,64 @@ resource "null_resource" "karpenter_nodepools_finalizer_cleanup" {
...
+  # unschedulable; finalize against a live node, issue #336). Idle 30 is half
+  # the jupyterlab default: an idle hour on the cheapest GPU node costs $0.53.
+  gpu_template_builtin = {
+    name         = "jupyterlab-gpu"
+    display_name = "JupyterLab GPU"
+    description  = "JupyterLab workspace with one NVIDIA GPU and persistent EBS storage"
+    gpus         = "1"
+    cpu          = "3500m"
+    memory       = "13Gi"
+    idle_minutes = "30"
```
earaghbidikashani: The built-in `jupyterlab-gpu` config hardcodes idle_minutes = "30". If an admin enables `enable_default_gpu_pool` and changes workspaces_idle_shutdown_timeout_min the plan fails with an error mentioning workspace_templates idle_minutes even though they never wrote one. 

Consider clamping the built-in idle to the deployment's window, or calling this out in the flag's docstring.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
earaghbidikashani, line 20: These still reference workspaces node groups / node_groups / instance_type / desired_size, none of which exist
 you fixed the identical stale reference in the "A workspace does not start" section just below as well

earaghbidikashani: same here

earaghbidikashani (approved):
overall LGTM, tiny comments

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
andrii-i: Good point, thanks @earaghbidikashani.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/workspaces.tf
```diff
@@ -85,17 +85,154 @@ resource "helm_release" "github_rbac" {
...
+      name             = name
+      isDefault        = "false"
+      displayName      = lookup(b.config, "display_name", name)
+      description      = lookup(b.config, "description", "JupyterLab workspace on the ${b.pools[0]} pool")
+      imageUri         = module.app_jupyterlab[0].image_uri
+      appType          = var.workspace_app_jupyterlab_app_type
+      accessType       = var.workspaces_default_access_type
+      ownershipType    = var.workspaces_default_ownership_type
+      storageClassName = local.workspace_storage_class
+      defaultResources = contains(keys(b.config), "cpu") ? {
```
bhanderig, line 179: The card pins `nvidia.com/gpu` to `min == max` here and the create page locks it, but a user can still POST a Workspace CR straight to the API asking for more GPUs than the card offers. Do we know the operator enforces `resourceBounds` on extended resources like `nvidia.com/gpu` at admission, not just cpu/memory? If it does, great — ignore this. If it doesn't, the only backstop is the pool's `max_gpus` ceiling, which caps the fleet but not a single workspace, so someone could pull a multi-GPU node off a 1-GPU card (and on a custom pool that omits `max_gpus`, nothing caps it). Worth confirming before we rely on the UI lock as the boundary.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
```diff
@@ -216,6 +216,64 @@ resource "null_resource" "karpenter_nodepools_finalizer_cleanup" {
...
+  }
+  # cpu/memory target the g4dn.xlarge allocatable (an over-pin is permanently
+  # unschedulable; finalize against a live node, issue #336). Idle 30 is half
+  # the jupyterlab default: an idle hour on the cheapest GPU node costs $0.53.
+  gpu_template_builtin = {
+    name         = "jupyterlab-gpu"
+    display_name = "JupyterLab GPU"
+    description  = "JupyterLab workspace with one NVIDIA GPU and persistent EBS storage"
+    gpus         = "1"
+    cpu          = "3500m"
```
bhanderig: This is pinned right at `g4dn.xlarge` allocatable, and since requests == limits there's basically no slack. If a future AL2023 AMI bumps kube/system reservations, or a cluster-wide DaemonSet (this plugin, CNI, EBS CSI) grows its requests, allocatable on the smallest node dips under `3500m`/`13Gi` and the pod never schedules — Karpenter brings up a node and the workspace just sits `Pending`. Validated on a live node today so I'm not blocking, but would you consider a little headroom (say `3` / `12Gi`) so reservation drift doesn't quietly break scheduling? The "finalize against a live node" note is a good flag to keep either way.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_components.py
```diff
@@ -242,6 +243,31 @@ def test_daemonset_component_status_ready(e2e_deployment: EndToEndDeployment) ->
         _poll_component_status(e2e_deployment, name, "Ready")
 
 
+@skip_if_testvars_not_set(["JD_E2E_GPU_ENABLED"])
+@pytest.mark.usefixtures("kubernetes_cluster_login")
+def test_nvidia_device_plugin_daemonset_when_gpu_enabled(e2e_deployment: EndToEndDeployment) -> None:
+    """The device-plugin DaemonSet exists when a GPU pool is configured.
+
+    Checked via kubectl rather than jd component: the GPU pieces are
+    deliberately absent from the manifest (components have no conditional
```
bhanderig, line 252: Makes sense that the device plugin can't go in the manifest yet. The thing I want to make sure we don't lose: when it's absent from `jd health`, a broken DaemonSet leaves every GPU workspace stuck `Pending` while health still reads green — and the only trail is the manual `kubectl get daemonset` step in TROUBLESHOOT. Is the "optional-component support" follow-up tracked in an issue somewhere? Just want it captured so this gap doesn't get forgotten.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
andrii-i: Added: `golden_karpenter_nodepools_cpu_gpu.yaml`, byte-compared like the CPU one (e1240e5).

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: That's deliberate, we need to build infra to deploy a real GPU cluster for testing + decide on when to run it (due to compute availability and cost constraints it's unreasonable to run it on every push in every PR). So planning to address it as follow-up outside of this PR, will create an issue.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
andrii-i: Went one step further than clamping: the built-in config now inherits `workspaces_idle_shutdown_timeout_default` (60 stock), so one knob governs every card and a window change can never fail the plan on a config you never wrote (b865376). A companion precondition also validates the default itself inside the window, with the error naming the actual variables (f2a9be3).

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/TROUBLESHOOT.md.template
andrii-i: Fixed (e6154d5).

andrii-i: Fixed together with the quota section above (e6154d5).

andrii-i:
Thank you for reviews @earaghbidikashani, @bhanderig, good comments and suggestions. I've addressed or replied to all of them. Please give this another look.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/workspaces.tf
andrii-i: Confirmed at admission, not just the UI. Checked the operator's validating webhook (`resource_validator.go` checks `resourceBounds` for every declared key, extended resources included) and verified it live on the test deployment: a server dry-run of a workspace requesting 2 GPUs against the 1/1 template was denied with `nvidia.com/gpu request 2 exceeds maximum 1 allowed by template`. `max_gpus` is indeed only the fleet cap; the per-workspace boundary is admission.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/platform_karpenter.tf
andrii-i: Took your numbers: `3`/`12Gi` (65a1de5), with the comment now stating the headroom rationale.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_components.py
andrii-i: Good point, created #344 to track this.

earaghbidikashani (approved):

bhanderig (approved):
LGTM

PR #347, opened 2026-08-27
feat: jd proxy

JGuinegagne:
This PR adds `jupyter-deploy-client-proxy`, a standalone, cloud-blind PyPI package (zero AWS deps) that reverse-proxies `http://localhost:PORT` to a remote self-signed-TLS endpoint and injects a short-lived, rotating auth token on every request. Its whole contract is the exec-credential-plugin pattern (similar to `aws eks get-token`): run a configured command, receive a JSON bundle `{host, port, ca_cert, headers, expires_at}`, connect over pinned TLS, and re-exec the command before expiry. It knows nothing about any cloud, and calls back to `jd` commands to rotate its auth header.

Also in this PR:
- `jupyter-deploy[proxy]` optional extra pins the package; `jd` only shells out to the console script, never imports it
- monorepo wiring: workspace member/source, dev deps (`trustme`, `types-aiofiles`), `aiologger`/`psutil` mypy overrides, and a `tests/functional` pytest split

### User experience
- a new `jupyter-deploy-client-proxy` console script tunnels `http://localhost:PORT` to any self-signed-TLS + bearer-token endpoint. It point it at `--token-command` (+ optional static `--ca-cert`) and it self-heals across credential rotation and network blips, no restart needed.

### Implementation
1. new workspace package `libs/jupyter-deploy-client-proxy` (hatchling, zero AWS/`jupyter_deploy` deps); console script → `cli/app.py` (typer): `--token-command`, `--listen-port`, `--ca-cert`, `--refresh-margin-seconds`, `--log-dir`, `--log-level`
2. `server/proxy.py` `JupyterDeployClientProxy`: aiohttp loopback server HTTP-parses each request, injects the bundle headers (incl. the WS upgrade), forwards over the pinned-TLS session; on `101` → transparent byte relay
    2.1. echoes the negotiated WS subprotocol on both legs (JupyterLab kernels speak `v1.kernel.websocket.jupyter.org` binary framing)
3. it owns the refresh loop + lifecycle (`start`/`stop`) and tracks a `ProxyState` (`enums.py`: `STARTING→RUNNING→DEGRADED`, `FAILED`); rebuilds the upstream session only on CA change, so the loopback port stays put
4. `credentials/`: pydantic `ConnectBundle` (`bundle.py`); exec-credential loop (`credential.py`) runs the token command, validates stdout, backs off exponentially; typed retryability (`exceptions.py`): timeout / `EX_TEMPFAIL` (75) / bad-JSON retryable, missing-binary / wrong-shape not
5. `server/tls.py`: client `SSLContext` pinned via `load_verify_locations(cadata=ca_cert)`, `check_hostname=False`, ALPN `http/1.1`: pin on the cert, not the address
6. `server/state.py`: versioned `status.json` written async via `aiofiles` (`{schema_version, state, pid, port, expires_at, process_created_at}`): the wire contract a later `jd proxy status` reads
7. `utils.py`: hop-by-hop/handshake header filtering with case-insensitive bundle-header merge; redacted bundle log summary (header names only)
8. `logger/`: `aiologger` wrapper (fire-and-forget with `flush`/`close`) + size-based rotation with out-of-band file/dir deletion recovery; file under `--log-dir`, else stderr
9. CLI `[proxy]` extra in `<jd>/pyproject.toml` pins `jupyter-deploy-client-proxy>=0.1.0`; the CLI stays template-agnostic and only shells out to the console script
10. root `pyproject.toml` + `uv.lock`: workspace member/source, dev deps, mypy overrides, and `norecursedirs` for `tests/functional` (run separately from the root unit suite)

### Testing
- unit: `tests/unit/` across bundle / credential (typed retry via counting scripts + exit codes) / utils / config / tls / rotation / logger / state / status / proxy / app; ~97% package coverage
- functional: `tests/functional/` spins a `trustme` self-signed aiohttp origin (HTTP + WS echo) driven in-process by the real proxy — HTTP forwarding, WS relay + subprotocol negotiation, TLS pin-mismatch refusal, credential-refresh re-exec, logging (file / rotation / stderr / deletion recovery), status reporting, and a console-script subprocess smoke; run via `just functional-test`
- manually with follow-up PRs that include `jd` CLI wiring (on `jd open` and w/ new `jd proxy` commands group) and a new `jupyterlab` template.


bhanderig (approved):

JGuinegagne:
> ## roborev: Combined Review (`3339ea3`)
> ## Verdict: Two medium-severity issues in the new `jupyter-deploy-client-proxy` — a CSRF/credential-injection surface on the loopback listener and a broken error path for post-prepare stream failures. No critical/high issues found.
> ### Medium
> 
> **1. Loopback proxy injects credentials and forges `Origin`/`Referer` with no request-side origin check — CSRF surface** `libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/utils.py:33-51` (`_rewrite_origin_headers` / `get_forwarded_request_headers`), interacting with `server/proxy.py` (`_forward_http` / `_relay_ws`)
> 
> For every incoming request the proxy injects the bundle's headers (including `Authorization: Bearer …`) and unconditionally rewrites any `Origin`/`Referer` to the upstream origin. While `jd proxy` runs, a malicious page in the user's browser can issue cross-origin state-changing requests to `http://127.0.0.1:<port>/...` (e.g. `POST /api/contents`, kernel-execute). Browsers send such requests even though they block reading the response — sufficient for CSRF. The proxy then injects the bearer credential the page could never obtain and rewrites `Origin`/`Referer` to pass the server's same-origin/XSRF check, defeating both defenses and yielding drive-by authenticated access (read/write files, start kernels, execute code).
> 
> _Mitigating factors:_ listener bound to `127.0.0.1`, port ephemeral by default — but `--listen-port` allows a fixed, guessable, browser-scannable port.
> 
> _Remediation:_ Add an origin allowlist on the proxy's own listener before injecting credentials/forwarding — reject requests whose `Origin`/`Referer` (or `Sec-Fetch-Site: cross-site`) don't correspond to the loopback listener itself, rather than normalizing them to the upstream origin. At minimum, document that the loopback listener is an unauthenticated credential-injection surface and gate cross-site requests.
> 
> **2. `ClientError` handler synthesizes a 502 even after the response has been prepared** `libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py:257-274` (`_forward_http`)
> 
> The `except aiohttp.ClientError` handler returns a fresh, unprepared `web.Response(status=502, ...)`, but a `ClientError` can be raised inside the `async for chunk in upstream.content.iter_any()` loop — after `await response.prepare(request)` has already sent status line + headers and possibly body chunks. Emitting a second response on an already-committed connection raises rather than degrading gracefully. This is a realistic path for long-lived/large streamed responses (e.g. `/api/contents` downloads, slow endpoints — the case the timeout constants call out) when the upstream drops mid-transfer.
> 
> _Fix:_ Distinguish pre-prepare from post-prepare failures. Only synthesize the 502 if `response` was never prepared; once prepared, abort the connection (break out and let `write_eof()`/connection close signal truncation) instead of constructing a new response.
> 

will address both findings in #353

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/utils.py
```diff
@@ -0,0 +1,87 @@
...
+def _rewrite_origin_headers(headers: dict[str, str], origin_host: str, origin_port: int) -> None:
+    """Rewrite ``Origin``/``Referer`` in place to the upstream origin.
+
+    The browser sends the loopback listener's origin (e.g. ``http://127.0.0.1:PORT``), which never
+    matches the instance-IP ``Host`` the request is forwarded to. Presenting the upstream origin lets
+    the server run its native same-origin/XSRF check instead of the template loosening ``allow_origin``.
+    Only rewrites headers already present (a missing ``Origin`` is same-origin by default upstream);
+    matches case-insensitively and preserves the original key casing.
+    """
+    origin = f"https://{origin_host}:{origin_port}"
```
bhanderig, line 40: **[Security]** The rewrite is unconditional — any present `Origin` becomes the upstream's, with no check that it *was* the loopback origin. It's applied on both legs (`proxy.py:237` HTTP, `proxy.py:275` WS upgrade).

The WS path is the concrete one. While the tunnel is up, any page in the user's browser can open:

```js
new WebSocket('ws://127.0.0.1:PORT/api/kernels/<id>/channels')
```

Browsers apply no CORS or preflight to WS handshakes, so the request lands; the proxy then injects the bundle's auth header and rewrites `Origin: https://evil.example` to the upstream origin.

Checking what's actually downstream in the base template: `jupyter-start.sh:19-20` runs Jupyter with `--IdentityProvider.token=` (auth fully delegated to the layer above) and `jupyter_server_config.py` never sets `allow_origin`, so Jupyter is on its default Origin-matches-Host check. XSRF does not apply to WebSocket upgrades at all — so on the kernel-channel endpoint that Origin check is the *sole* cross-origin defense, and we rewrite it so it always passes.

The docstring's premise is correct (the template genuinely does not loosen `allow_origin`), but this trades a scoped config change for an unscoped bypass. Suggest rejecting before rewriting, in `_handle` so it covers both legs:

```python
origin = request.headers.get("Origin")
if origin is not None and origin not in {f"http://127.0.0.1:{self._port}", f"http://localhost:{self._port}"}:
    return web.Response(status=403, text="cross-origin request rejected")
```

`test_utils.py:53` only feeds a loopback origin, so a foreign-origin case is worth adding either way.

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    return f"{bundle.host}:{bundle.port} headers={sorted(bundle.headers)} expires_at={bundle.expires_at.isoformat()}"
+
+
+def get_seconds_until_refresh(expires_at: datetime, margin_seconds: float = DEFAULT_REFRESH_MARGIN_SECONDS) -> float:
+    """Return how long to sleep before re-execing the token command.
+
+    Refreshes ``margin_seconds`` before ``expires_at``; never negative.
+    """
+    delta = (expires_at - datetime.now(UTC)).total_seconds() - margin_seconds
+    return max(0.0, delta)
```
bhanderig, line 78: **[Reliability]** `max(0.0, delta)` means a bundle that arrives already inside `refresh_margin_seconds` yields a zero sleep, and the refresh loop re-execs the token command immediately — that command being `jd proxy connect-info`, i.e. cloud API calls at subprocess speed. Triggers: client clock skew ahead of the issuer, or a credential TTL shorter than the 15s default margin.

Worth noting an "expiry advanced since last fetch" guard would *not* catch this — with a 10s TTL and a 15s margin every fetch does return a later `expires_at` (wall clock moved forward), so the guard passes while the loop still spins. Two clamps do hold:

```python
remaining = (expires_at - datetime.now(UTC)).total_seconds()
delay = remaining - margin_seconds
if delay < MIN_REFRESH_SLEEP_SECONDS:
    # margin >= remaining lifetime: refresh at a fraction of what's left, not immediately
    delay = max(MIN_REFRESH_SLEEP_SECONDS, remaining * REFRESH_LIFETIME_FRACTION)
return delay
```

With `1.0` / `0.5` in `constants.py`, a 10s credential refreshes every ~5s instead of continuously. Suggest also logging at `warning` rather than `debug` at `proxy.py:206` when the clamp engages — that state means `--refresh-margin-seconds` is wrong for the credential, and it's currently invisible.

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
```diff
@@ -0,0 +1,325 @@
...
+            return await self._relay_ws(request)
+        return await self._forward_http(request)
+
+    async def _forward_http(self, request: web.Request) -> web.StreamResponse:
+        assert self._session is not None and self._bundle is not None
+        url = f"https://{self._bundle.host}:{self._bundle.port}{request.raw_path}"
+        headers = get_forwarded_request_headers(
+            request.headers, self._bundle.headers, self._bundle.host, self._bundle.port
+        )
+        body = await request.read()
```
bhanderig, line 240: **[Reliability]** `web.Application()` at line 96 takes aiohttp's default `client_max_size=1048576`, and `request.read()` is the code path that *enforces* it — so this 413s rather than merely buffering. Repro using those exact two lines (`web.Application()` + `await request.read()`):

```
PUT  0.50 MiB -> HTTP 200
PUT  2.00 MiB -> HTTP 413
```

The 413 comes from the proxy itself; the request never reaches upstream. On `/api/contents` — how JupyterLab both saves and uploads — bodies are JSON with binary content base64-encoded, so payload is roughly 4/3 x file size plus envelope: a notebook with a couple of embedded plot outputs, or a single-shot upload over ~768 KiB, fails.

One line fixes the ceiling *and* the buffering, since streaming never calls `read()`:

```python
# `request.read()` both materializes the whole body and enforces aiohttp's 1 MiB
# `client_max_size`, which 413s notebook saves and uploads. `can_read_body` keeps
# bodyless requests (GET) from being framed as chunked.
data = request.content if request.can_read_body else None
```

Tradeoff: `content-length` is in `DROP_FROM_REQUEST_HEADERS`, so a streamed body goes upstream as `Transfer-Encoding: chunked`. Tornado and Traefik both accept that; if you'd rather keep length framing, forward `content-length` when present and aiohttp will honor it with the stream.

A functional test that PUTs ~2 MiB and asserts a 200 is about three lines in the existing harness, and is the assertion that would have caught this.

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
```diff
@@ -0,0 +1,325 @@
...
+                self._logger.debug(f"{request.method} {request.path} -> {upstream.status}")
+                response = web.StreamResponse(
+                    status=upstream.status, headers=get_forwarded_response_headers(upstream.headers)
+                )
+                await response.prepare(request)
+                async for chunk in upstream.content.iter_any():
+                    await response.write(chunk)
+                await response.write_eof()
+                return response
+        except aiohttp.ClientError as e:
```
bhanderig, line 254: **[Reliability]** This `except` also covers the streaming loop that runs *after* `response.prepare(request)` has put status and headers on the wire — e.g. `ServerTimeoutError` from the deliberate `sock_read=300`, or a mid-response disconnect, which the design explicitly expects since the upstream IP "changes across host stop/start". Returning a fresh `web.Response` at that point writes a second HTTP message into the live connection:

```
HTTP/1.1 200 OK
Transfer-Encoding: chunked
...
b
{"partial":
HTTP/1.1 502 Bad Gateway          <-- second response, inside the first's chunked body
Content-Type: text/plain; charset=utf-8
Content-Length: 26
...
1a
upstream unreachable: boom
0
```

So the client receives a **200** whose body contains a literal HTTP response, with mixed chunk framing. Suggest splitting the pre- and post-commit cases — once headers are out, aborting is the only honest signal:

```python
except aiohttp.ClientError as e:
    if response is None:
        self._logger.warning(f"upstream unreachable for {request.method} {request.path}: {e}")
        return web.Response(status=502, text=f"upstream unreachable: {e}")
    # Headers are already on the wire — a 502 here lands *inside* this response's body.
    self._logger.warning(f"upstream failed mid-response for {request.method} {request.path}: {e}")
    if request.transport is not None:
        request.transport.abort()
    return response
```

(with `response: web.StreamResponse | None = None` hoisted above the `try`).

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
```diff
@@ -0,0 +1,325 @@
...
+                # A burst of attempts with backoff (the refresh budget, larger than startup's);
+                # this loop keeps retrying cycles forever.
+                delay = get_seconds_until_refresh(
+                    self._bundle.expires_at, margin_seconds=self._config.refresh_margin_seconds
+                )
+                self._logger.debug(f"next credential refresh in {delay:.0f}s")
+                await asyncio.sleep(delay)
+                bundle = await self._fetch_bundle(self._config.refresh_max_attempts)
+                await self._apply_bundle(bundle)
+            except TokenCommandError:
```
bhanderig, line 210: **[Observability]** `TokenCommandError` is the base of both subclasses (`exceptions.py:18-23`), so `NotRetryableTokenCommandError` — missing binary, wrong bundle shape — lands in DEGRADED and gets re-attempted every 30s indefinitely.

That inverts what this PR documents in three places:
- `fetch_bundle_with_retries`: "Non-retryable failures raise immediately"
- `_fetch_bundle`, lines 161-162: "permanent failures (missing binary, bad bundle shape) **raise immediately**"
- `ProxyState`: DEGRADED is "refresh is failing"; FAILED is for when "the refresh machinery is dead and won't self-heal"

The typed-retryability work (EX_TEMPFAIL, exit-code branching, its unit tests) is built carefully and then discarded at the one call site where the distinction changes behavior. `start()` gets it right — a NotRetryable propagates to `except Exception` at line 109 → FAILED → re-raise — so the asymmetry is only here. It's also untested: `test_proxy.py` covers the FAILED transition when `_apply_bundle` raises, but has no `TokenCommandError` case at all, so nothing separates the two branches.

Concretely: a rebuilt venv takes `jd` off PATH, and the proxy re-execs a missing binary every 30s forever, serving a credential that expires underneath it, reporting "degraded, refresh is failing" when the honest answer is "restart me."

```python
except NotRetryableTokenCommandError as e:
    # Permanent — retrying is pointless, and DEGRADED promises self-healing that cannot happen.
    self._logger.error(f"refresh permanently broken, stopping refresh: {e}")
    self._state = ProxyState.FAILED
    await self.write_status_best_effort()
    break
except RetryableTokenCommandError:
    ...  # existing DEGRADED path
```

plus a test per branch. If DEGRADED-forever is actually the intent, then the fix is in the docstrings instead — the problem is that code and docs currently disagree.

pyproject.toml
```diff
@@ -135,4 +156,9 @@ markers = [
   "bare: Tests requiring the bare install track with no AWS deps (select with '-m bare')",
 ]
 addopts = "-m 'not e2e' --import-mode=importlib"
-norecursedirs = ["**/tests/e2e"]
+# tests/functional run via `just functional-test` (prepend import-mode, per-package rootdir) — not here.
+norecursedirs = ["**/tests/e2e", "**/tests/functional"]
```
bhanderig, line 160: **[Review-risk]** These 17 functional tests run in no CI lane, and structurally can't:

```console
$ uv run .github/utils/get_modified_dirs.py "main"
["."]
$ uv run pytest --collect-only -q          # exactly what ci.yml runs
0 functional tests collected               # (106 unit collected)
```

Three things compound: `get_modified_dirs.py` `LIB_PATHS` doesn't list `libs/jupyter-deploy-client-proxy`; any file outside those paths sets `root_changed` and collapses the matrix to `["."]` (this PR touches `AGENT.md`, `justfile`, `pyproject.toml`, `uv.lock`); and at root this `norecursedirs` excludes them, while `test.yml` runs a bare `uv run pytest` and nothing anywhere invokes `just functional-test`.

The structural part: because the package isn't in `LIB_PATHS`, even a future PR touching *only* proxy files falls into the `root_changed` branch and gets `["."]` again — so the package's own `testpaths` and its `fail_under = 80` gate never execute in CI either. TLS pin-mismatch refusal, WS subprotocol negotiation, credential-refresh re-exec, rotation, the console-script smoke: all developer-discretion only.

Same objection raised on #341 ("`JD_E2E_GPU_ENABLED` is set in no committed workflow that I see... skip in every automated lane"), but without the cost defense that justified deferring it there — this suite is hermetic (trustme + in-process aiohttp, no cloud, no Docker).

Fix needs both parts, since the matrix keeps collapsing on any root-touching PR: add `libs/jupyter-deploy-client-proxy` to `LIB_PATHS`, and add an unconditional `functional-test` job to `ci.yml` running `uv run pytest -c pyproject.toml --import-mode=prepend libs/jupyter-deploy-client-proxy/tests/functional`.

If the split exists only because `harness` needs `--import-mode=prepend`, then making the harness importable (a `tests/functional/__init__.py`, or a conftest fixture) and dropping the split entirely is less CI surface than a dedicated job.

Adjacent, pre-existing: `libs/jupyter-deploy-tf-aws-eks-oidc` is also missing from `LIB_PATHS`.

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
```diff
@@ -0,0 +1,325 @@
...
+        # channels request `v1.kernel.websocket.jupyter.org`, whose framing is binary; if the
+        # downstream response doesn't echo the negotiated subprotocol back to the browser, the
+        # client silently falls back to the v0 text protocol while the server speaks v1, and
+        # every kernel message fails to deserialize ("cannot convert 'str' object to bytes").
+        # Passing `protocols` to both ends lets aiohttp regenerate the handshake header (the raw
+        # sec-websocket-protocol header is dropped in get_forwarded_request_headers).
+        client_protocols = [
+            p.strip() for p in request.headers.get("Sec-WebSocket-Protocol", "").split(",") if p.strip()
+        ]
+        downstream = web.WebSocketResponse(protocols=client_protocols)
```
bhanderig, line 271: `non-blocking:` three smaller items, grouped to keep the thread count down.

**1. WS subprotocol is committed downstream before upstream negotiates (this line).** `downstream.prepare()` runs here, before `ws_connect` at line 280. Since the server-side `protocols` list is the client's own list, aiohttp picks the client's **first** preference and sends it in the 101 — before upstream has been asked. Upstream then chooses independently and nothing reconciles them.

Mostly latent today: current JupyterLab offers a single subprotocol, and I checked the base template's Traefik (static config plus forwardauth/error middlewares — nothing touches `Sec-WebSocket-Protocol`). The reachable half is the error path — when upstream *rejects* the upgrade, the `ClientError` handler at line 282 closes a socket that already completed a successful 101, so the browser sees connect-then-close rather than a failed upgrade and retries in a loop.

Reordering fixes both:

```python
downstream: web.WebSocketResponse | None = None
try:
    async with self._session.ws_connect(url, headers=headers, protocols=client_protocols) as upstream:
        downstream = web.WebSocketResponse(protocols=[upstream.protocol] if upstream.protocol else [])
        await downstream.prepare(request)
        await self._pipe_ws(downstream, upstream)
except aiohttp.ClientError as e:
    if downstream is None:
        return web.Response(status=502, text=f"ws upstream error: {e}")
    await downstream.close()
return downstream
```

Note `harness.py:44-46` echoes back whatever the client requested, so the fake origin always picks the client's first preference — the harness can't express this case. A `server_protocols` param would make it assertable.

**2. `[proxy]` extra will be unresolvable when it ships** (`libs/jupyter-deploy/pyproject.toml:56`). `jupyter-deploy-client-proxy` is 404 on PyPI, and this pin ships in the built wheel's metadata — `[tool.uv.sources]` is uv-only and stripped at build time. So `pip install "jupyter-deploy[proxy]"` is a hard resolution failure, and there's no `release-client-proxy.yml` alongside the four existing release workflows. (PyPI today: `jupyter-deploy` 0.6.8, `provides_extra: ["aws","k8s"]`.)

This is the first first-party sibling to be a runtime dep of the published CLI — the other two in `[tool.uv.sources]` appear in no `dependencies` list, and `pytest-jupyter-deploy` is published but depended on by nothing — so release ordering is newly load-bearing: publish the lib, then cut the CLI release. Also `>=0.1.0` has no upper bound on a dependency whose whole contract is a wire format this PR defines (`ConnectBundle`, `status.json` `schema_version`); `>=0.1.0,<0.2.0` costs nothing now. Is the release workflow tracked in an issue?

**3. Package boundary rules aren't documented or enforced** (`AGENT.md`). The new section covers how to run the tests, but the package gets no entry in `# Project Context`, where every other workspace member has one with its `Code:` path and boundary rules. The two invariants that define this package live only in a pyproject comment: zero cloud/`jupyter_deploy` deps, and **`jd` must never import it, only shell out to the console script**.

That second one is enforceable with a pattern already here — `tests/e2e/test_bare_installation.py` asserts one-import-per-extra (`[aws]` → boto3, `[k8s]` → kubernetes), and its docstring is explicit that a bare track quietly gaining deps "is a real bug." The new extra breaks that 1:1, and the bare track does run in CI (`smoke-bare`, plus release-cli's `variant` matrix):

```python
    def test_no_client_proxy(self) -> None:
        with self.assertRaises(ImportError):
            import jupyter_deploy_client_proxy  # noqa: F401
```

That single test is what mechanically enforces "never imported" — a future handler importing it turns the bare track red.

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
JGuinegagne: not sure this actually matters, as requests would get chunked up into smaller HTTP requests.

pyproject.toml
JGuinegagne: I know, for now we are running them manually; will add a separate job for them.

JGuinegagne:
Will address remaining comments in #353

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/utils.py
JGuinegagne: man LLMs can write a lot, in an intricate way, to say little.
Yes, the constants must be consistent, that's...obvious. I guess we can add a unit tests for the constant consistency...

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py
JGuinegagne: nvm actually, nice catch, I'll fix in #353

libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    ``Origin``/``Referer`` to the upstream origin so the server's native same-origin check passes."""
+    filtered = {k: v for k, v in incoming.items() if k.lower() not in DROP_FROM_REQUEST_HEADERS}
+    merged = _merge_bundle_headers_into_incoming(filtered, bundle_headers)
+    _rewrite_origin_headers(merged, origin_host, origin_port)
+    return merged
+
+
+def get_forwarded_response_headers(upstream: Mapping[str, str]) -> dict[str, str]:
+    """Filter hop-by-hop headers off an upstream response before relaying it downstream."""
+    return {k: v for k, v in upstream.items() if k.lower() not in DROP_FROM_RESPONSE_HEADERS}
```
andrii-i, line 64: `get_forwarded_response_headers` returns a plain `dict`, so when a response carries two or more response headers under the same name, only the last one reaches the browser. See the  `Set-Cookie` headers example: run `jupyter lab` in the template's config and read `/lab` with `curl -i`: it sets `username-<host>` and `_xsrf`; put the same response through the proxy and only `_xsrf` comes back, so the identity cookie is dropped and only `_xsrf` is kept. We should build the forwarded response headers as a multidict so repeated headers pass through: a multidict holds several values under one key, and aiohttp writes each one back as its own `Set-Cookie` line. Fine as a follow-up.

andrii-i (approved):
Thanks @JGuinegagne, the architecture looks good. One note below on response header of the same name being lost.

PR #348, opened 2026-08-27
feat: proxy cmds

JGuinegagne:
This PR adds the `jd proxy` command group and the proxy-mode `jd open` flow that drive the standalone client proxy from #347, which handles access to a project whose template exposes no public URL via a localhost tunnel. The connect-info bundle the proxy consumes is assembled entirely from manifest-declared instructions.

Depends on: #347

### User experience
- new `jd proxy` command group:
    - `connect-info`: used by the proxy to setup or refresh its auth header
    - `start`: create a proxy as a local process in background mode
    - `open`: open an existing proxy process in the web browser
    - `stop`: kills an existing proxy process
    - `status` / `show`: information about an existing proxy process
- `jd open`: on a proxy-mode template, deletes any running proxy, launches the local proxy as foreground, opens the loopback URL it binds. Since the proxy runs as foreground, the `jd open` manages the proxy process, CTRL+C terminates both the `jd` command _and_ the proxy process.
— `jd open -d`: deletes any running proxy, launch the proxy as background it and returns; users can terminate the proxy with `jd proxy stop` afterwards
- actionable errors instead of tracebacks:
    - proxy extra not installed → install hint
    - a proxy already running → `stop`/`open` hints
    - no proxy running → `start` hint
- `jd proxy stop` never kills a recycled PID: it leaves an alive-but-unconfirmed process untouched with a pointer to its runtime dir

### Implementation
1. `cli/proxy_app.py`: the `jd proxy` typer group; each command resolves the project, delegates to `ProxyHandler`, and displays; all orchestration lives in the handler
2. `handlers/proxy_handler.py`: `ProxyHandler.get_connect_bundle()` runs the manifest `proxy.connect-info` sequence via the generic command runner → `ProxyConnectBundle`; delegates process lifecycle to a private `ProxyManager`
3. `proxy/proxy_manager.py`: `ProxyManager` owns setting the proxy runtime dir (`.jd-proxy/<kind>/<name>/<ts>/`), launch (attached/detached), wait-for-bind, browser open, and confirmed teardown; no manifest/engine/cloud deps
    3.1. `resolve_console_script` invokes the `jupyter-deploy`/`jupyter-deploy-client-proxy` scripts by absolute path, so that launch + token re-exec don't depend on the child's `PATH`
    3.2. `_announce_and_open` refuses any non-loopback URL before opening the browser (a manifest `path` can't smuggle a host via userinfo)
4. `proxy/proxy_utils.py` + `handlers/payloads.py`: versioned `status.json` reader (tolerant of missing/corrupt, propagates a present-but-unreadable file) + `ProxyConnectBundle`/`ProxyStatus`
5. `manifest.py`: `open: {mode, path}` declaration (`get_open`) and `supports_proxy()`, gated on the presence of the `proxy.connect-info` command
6. AWS connect-info instructions:
    6.1. `api/aws/ec2/ec2_instance` (live public-IP resolve), `api/aws/ec2/ec2_security_group` (reconcile caller `/32`)
    6.2. `api/aws/ssm/ssm_parameter` (read cert pin)
    6.3. wired through `aws_ec2_runner`/`aws_ssm_runner`
7. `provider/aws/aws_sts_runner.py`: `mint-connect-token` builds the `k8s-aws-v1` token + the opaque `Authorization`/`x-k8s-aws-id` header map; `api/aws/sts/eks_token.py` generalized `cluster_name`→`binding_id` (k8s call site updated)
8. `handlers/project/open_handler.py` + `cli/app.py`: proxy-mode `jd open` drives a `ProxyManager` directly (replace-then-open, `--detached`, `wait()`); `_is_secure_open_url` now accepts `http` loopback
9. `cmd_utils.py` (+ `psutil` as a core CLI dep): pid handler and identity guards behind stop/replace/status
10. `exceptions.py` + `cli/error_decorator.py`: typed proxy exceptions rendered with hints; `cli/{simple,progress}_display.py` gain `set_status` for spinner-only phase narration

### Testing
- unit: covers the new surface, most importantly `test_proxy_app` (all six commands, including handler-failure propagation), `test_proxy_manager` (lifecycle, console-script resolution, loopback-URL guard), `test_open_handler` (proxy delegation + loopback gate)
- manual: `jd proxy start`/`open`/`status`/`show`/`stop` and `jd open` against a live deployed proxy-mode template. The loopback tunnel reaches the app end-to-end; also verified the `jd`-by-absolute-path / venv-not-on-`PATH` invocation now works (3.1)

### Screenshots
**jd proxy --help**
Screenshot 2026-08-27 at 2 17 39 PM

**jd proxy start --help**
Screenshot 2026-08-27 at 2 18 02 PM

**jd proxy start**
Screenshot 2026-08-27 at 2 22 17 PM

**jd open --help**
Screenshot 2026-08-27 at 2 23 04 PM

**jd open**
Screenshot 2026-08-27 at 2 24 10 PM

**jd open -d**
Screenshot 2026-08-27 at 2 25 34 PM

And finally:
**app**
Screenshot 2026-08-27 at 2 24 33 PM


JGuinegagne:
> ## roborev: Combined Review (`7714001`)
> ## Code Review: 1 Medium finding (all other findings below threshold)
> ### 🟡 Medium
> 
> **Forwarded requests capped by aiohttp's default 5-minute timeout** — `libs/jupyter-deploy-client-proxy/jupyter_deploy_client_proxy/server/proxy.py:1215` (`_apply_bundle`), `:1270-1282` (`_forward_http`)
> 
> The upstream `aiohttp.ClientSession` is created with no `timeout` override, so aiohttp's default `ClientTimeout(total=300)` applies to every forwarded request. Because `total` bounds the _entire_ operation including reading the streamed response body, any response proxied for more than 5 minutes (large file download via `/api/contents`, a slow/streamed endpoint) is severed mid-stream, surfacing to the browser as a truncated download or a 502.
> 
> **Fix:** Pass an explicit `timeout=aiohttp.ClientTimeout(total=None, sock_connect=..., sock_read=...)` (or at least a large/disabled `total`) when constructing the `ClientSession`, so request duration isn't capped by the client library.
> 
> _Both reviewers noted the change shows careful security practice overall (loopback-only binding, redacted log summaries, userinfo-smuggling guard on browser-open, PID-identity confirmation before signaling, TLS pinning, argv-list `subprocess` calls). Remaining findings were below the Medium threshold and are omitted._
> 

#347 scope

uv.lock
```diff
@@ -709,6 +709,56 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/57/30/4a22984d4f1bdfb8c054f07a92bc176b97a3134cc1d6c4b3bffb1f3688b4/cryptography-50.0.0-cp39-abi3-win_amd64.whl", hash = "sha256:d24fead1d4d076e1bfb006dcec392074a3cd8d7b4fc8a595aa64073b2b7a96ba", size = 3874135, upload-time = "2026-07-31T14:24:50.085Z" },
 ]
 
+[[package]]
+name = "cryptography"
```
andrii-i, line 663: Seems like this is a duplicate `cryptography` block from the `main` merge, and `build` errors out on it. Regenerate the lock file to remove it.

libs/jupyter-deploy/jupyter_deploy/constants.py
```diff
@@ -3,6 +3,7 @@
 VARIABLES_DEFAULTS_FILENAME = "variables-defaults.yaml"
 HISTORY_DIR = ".jd-history"
 JD_DIR = ".jd"
+PROXY_RUNTIME_DIR = ".jd-proxy"
```
andrii-i, line 6: `.jd-proxy/` isn't in the [`.gitignore` that `jd init` writes](https://github.com/jupyter-infra/jupyter-deploy/blob/7bb9e1b189700b87eb9be12cc0e076b81d4d635a/libs/jupyter-deploy/jupyter_deploy/handlers/init_static/gitignore/.gitignore.terraform.template), and the S3 backup on `jd up`/`jd down` [excludes files by that `.gitignore`](https://github.com/jupyter-infra/jupyter-deploy/blob/7bb9e1b189700b87eb9be12cc0e076b81d4d635a/libs/jupyter-deploy/jupyter_deploy/provider/aws/store/s3_store.py#L116), so it uploads the proxy runtime dirs with their logs (up to 5 x 10 MiB per start, and old dirs are never cleaned up). We should add `.jd-proxy/` next to `.jd-history/` in the template.

libs/jupyter-deploy/jupyter_deploy/proxy/proxy_manager.py
```diff
@@ -0,0 +1,483 @@
...
+# Default target for single-app templates (e.g. aws:ec2:jupyterlab).
+DEFAULT_TARGET_KIND = "server"
+DEFAULT_TARGET_NAME = "default"
+
+# How long to wait for a launched proxy to bind + publish its port.
+_LISTENING_TIMEOUT_SECONDS = 10.0
+_LISTENING_POLL_INTERVAL_SECONDS = 0.1
+
+# How long to wait for the app to answer through the proxy (SG door reconcile + upstream
+# TLS handshake + app readiness) before opening the browser anyway.
```
andrii-i: This phase doesn't exist anymore. Should this be removed?

andrii-i (dismissed):
Thanks @JGuinegagne, this looks good. A couple of non-blocking comments below.

libs/jupyter-deploy/jupyter_deploy/constants.py
JGuinegagne: nice catch, thanks!

andrii-i (approved):
Thank you, @JGuinegagne.

PR #330, opened 2026-08-03
feat: combined pool

JGuinegagne:
This PR unifies EKS Managed Node Groups (MNG) and Karpenter NodePools under one `jd pool list/show/status` interface, so users inspect every pool of hosts blind to which subsystem backs it. All pool-type knowledge stays in the template; the handler remains provider-agnostic, which is the constraint the first attempt (reverted in #322) violated.

Also in this PR:
- bump `eks-oidc` chart dependencies: karpenter `1.3.8`→`1.13.1` (required for k8s 1.36 — the old pin logged `not compatible with K8s version 1.36`), keda `2.16.1`→`2.20.2`, prometheus `27.0.0`→`29.21.0`, cluster-autoscaler `9.58.0`→`9.59.0`, operator `0.2.0`→`0.2.1`, router `0.1.1`→`0.2.0`; grant the karpenter controller `iam:ListInstanceProfiles` (its GC reconciler needs it from v1.7+)
- rename the platform MNG to a static `platform` so it reads cleanly in `jd pool list` (was `${cluster_name}-platform`)
- refresh the stale `eks-oidc` Helm charts doc table (listed 3 of the 11 charts actually deployed)

Closes: #324
Closes: #319

### User experience
- `jd pool list` returns every pool — the `platform` MNG **and** the Karpenter NodePools (`routing`, `workspace-cpu`) — as one flat name list, MNG first
- `jd pool show --name platform` / `jd pool status --name karpenter` work for either kind:
  - a Karpenter pool returns the NodePool CRD; the `platform` MNG returns the EKS `DescribeNodegroup` blob
  - status normalizes to the same `Ready`/`Creating`/`Degraded` vocabulary across both
- an unknown pool name fails gracefully (clean not-found, no traceback); so does `jd pool list` on an initialized-but-undeployed project

### Implementation
1. new built-in combinator runner in `<jd>/provider/core/core_runner.py`, routed via a reserved `core` ApiGroup (`<jd>/provider/enum.py`, `instruction_runner_factory.py`)
    1.1. `concat-lists-of-str` concatenates `list[str]`; missing/skipped coerces to `[]`
    1.2. `coalesce-str` returns the first non-empty `str` operand
    1.3. type-suffixed names leave room for future `-of-int` variants;
2. command-level `flags`/`conditions` + step-level `when:` schema in `<jd>/manifest.py`
    2.1. a flag is a named boolean = its conditions ANDed; `when: "!flag"` negates
    2.2. pydantic validators reject a malformed `when:`
[...]

PR #350, opened 2026-08-27
fix: allow karpenter to read its generated instance profiles

### LINKED ISSUE #349: Deleting a workspace pool leaves its EC2NodeClass stuck in Terminating
## Problem

Removing or renaming a workspace pool (for example turning `enable_default_gpu_pool` off, or editing a `workspace_nodepools` entry) deletes the pool's EC2NodeClass, and the deletion never completes: the object stays in `Terminating` with the `karpenter.k8s.aws/termination` finalizer indefinitely. Re-creating a pool under the same name then collides with the terminating object: the new NodePool reports `NodeClassReady=False`, and every workspace scheduled to that pool stays Pending.

The chart runs EC2NodeClasses with a pre-created instance profile (`spec.instanceProfile`), which makes sense as it keeps all IAM writes at deploy time in terraform instead of granting them to the controller. But Karpenter's termination reconciler still probes the `<cluster>_<hash>` instance profile names it would have auto-managed: on NodeClass deletion it calls `iam:GetInstanceProfile` for those names and [treats NotFound as already clean](https://github.com/aws/karpenter-provider-aws/blob/v1.13.1/pkg/providers/instanceprofile/instanceprofile.go#L188-L194). The controller policy [allows `iam:GetInstanceProfile` only on the single pre-created profile](https://github.com/jupyter-infra/jupyter-deploy/blob/1617d76046561e4ec0af340ef29dfa51fb75cbba/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf#L356-L360), so the probe on the generated names returns `AccessDenied` instead of `NoSuchEntity`, and the finalizer retries forever.

## Reproduction

Deploy the eks-oidc template with `enable_default_gpu_pool: true`, turn the flag off, and run `jd config` and `jd up`: the pool's EC2NodeClass stays in `Terminating` with the `AccessDenied` above on every reconcile. Turning the flag back on leaves the new `workspace-gpu` NodePool `NodeClassReady=False` and its workspaces Pending.

## Proposed solution

Add the cluster-prefixed generated names to the `AllowInstanceProfileGet` statement: `arn:<partition>:iam::<account>:instance-profile/<cluster>_*` alongside the pre-created profile ARN. With the read allowed on those names, IAM would no longer reject the probe before it can look anything up: the lookup finds no such profile and returns `NoSuchEntity`, which the reconciler treats as nothing to clean up, so the finalizer clears and deletion completes. The added permission is read-only.

Alternatives to matching the generated names with an ARN pattern would be to:
- Scope the read by resource tags instead of by name: tag the profiles Karpenter creates and allow reading any profile carrying that tag. This cannot work here. To decide whether such a read is allowed, IAM has to look at the tags of the profile being read. But the profile Karpenter asks about does not exist, which is exactly what Karpenter is trying to confirm, and a profile that does not exist has no tags. So IAM cannot allow the read and answers AccessDenied, and Karpenter is back in the same stuck state.
- Allow `iam:GetInstanceProfile` on every instance profile in the account (`instance-profile/*`), so that any name the reconciler asks about is readable and a missing profile always comes back as `NoSuchEntity`, as both the upstream Karpenter reference policy and the terraform-aws-modules Karpenter policy do. Deletion completes exactly as with the cluster-prefixed grant, but anyone who compromises the controller can then look up any instance profile in the account by name and see the IAM role it carries; the cluster prefix limits the read to the names Karpenter generates for this cluster.

andrii-i:
Fixes #349.

Adds the cluster-prefixed generated names to the `AllowInstanceProfileGet` statement: `arn:<partition>:iam::<account>:instance-profile/<cluster>_*` alongside the pre-created profile ARN. With the read allowed on those names, IAM would no longer reject the probe before it can look anything up: the lookup finds no such profile and returns `NoSuchEntity`, which the reconciler treats as nothing to clean up, so the finalizer clears and deletion completes. The added permission is read-only.

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
@@ -353,10 +353,17 @@ data "aws_iam_policy_document" "karpenter_controller" {
     }
   }
 
+  # Must also cover the <cluster>_<hash> profile names Karpenter auto-generates:
+  # the EC2NodeClass termination reconciler probes them on every delete even in
+  # pre-created-profile mode, and a 403 there (vs the expected 404) blocks the
+  # finalizer, wedging the NodeClass in Terminating forever.
   statement {
-    sid       = "AllowInstanceProfileGet"
-    actions   = ["iam:GetInstanceProfile"]
-    resources = [aws_iam_instance_profile.karpenter_node.arn]
+    sid     = "AllowInstanceProfileGet"
+    actions = ["iam:GetInstanceProfile"]
+    resources = [
+      aws_iam_instance_profile.karpenter_node.arn,
+      "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:instance-profile/${module.eks_cluster.cluster_name}_*",
+    ]
   }
 
   # Karpenter >=1.7 runs an instance-profile garbage-collection reconciler that calls
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
@@ -353,10 +353,12 @@ data "aws_iam_policy_document" "karpenter_controller" {
     }
   }
 
-  # Must also cover the <cluster>_<hash> profile names Karpenter auto-generates:
-  # the EC2NodeClass termination reconciler probes them on every delete even in
-  # pre-created-profile mode, and a 403 there (vs the expected 404) blocks the
-  # finalizer, wedging the NodeClass in Terminating forever.
+  # Must also cover the <cluster>_<hash> profile names Karpenter auto-generates
+  # (EC2NodeClass.LegacyInstanceProfileName as of karpenter v1.13.1): the termination
+  # reconciler probes them on every delete even in pre-created-profile mode, and a 403
+  # there (vs the expected 404) blocks the finalizer, leaving the NodeClass stuck in
+  # Terminating. New-style profiles live under an IAM path (/karpenter/<region>/...)
+  # and are only touched by ListInstanceProfiles, granted below.
   statement {
     sid     = "AllowInstanceProfileGet"
     actions = ["iam:GetInstanceProfile"]
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
@@ -114,6 +114,21 @@ def test_all_eks_addons_gated_by_cluster_addons() -> None:
     )
 
 
+def test_karpenter_can_read_generated_instance_profiles() -> None:
+    """The AllowInstanceProfileGet statement MUST cover the <cluster>_* instance
+    profile names Karpenter generates: its EC2NodeClass termination reconciler
+    probes them with GetInstanceProfile, and a 403 there (vs the expected 404)
+    blocks the finalizer, wedging the NodeClass in Terminating forever.
+    """
+    iam_tf = (TEMPLATE_PATH / "engine" / "iam.tf").read_text()
+    statement = re.search(r'sid\s*=\s*"AllowInstanceProfileGet"(.*?)(?=\n  statement)', iam_tf, re.DOTALL)
+    assert statement is not None, "AllowInstanceProfileGet statement not found in iam.tf"
+    assert "instance-profile/${module.eks_cluster.cluster_name}_*" in statement.group(1), (
+        "iam:GetInstanceProfile no longer covers Karpenter's generated <cluster>_* "
+        "instance profiles — deleting an EC2NodeClass will wedge on a 403 probe."
+    )
+
+
 def _iter_resource_blocks(content: str) -> list[tuple[str, str, str]]:
     """Yield (resource_type, resource_name, body) for every resource block in content."""
     blocks: list[tuple[str, str, str]] = []
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
@@ -115,17 +115,15 @@ def test_all_eks_addons_gated_by_cluster_addons() -> None:
 
 
 def test_karpenter_can_read_generated_instance_profiles() -> None:
-    """The AllowInstanceProfileGet statement MUST cover the <cluster>_* instance
-    profile names Karpenter generates: its EC2NodeClass termination reconciler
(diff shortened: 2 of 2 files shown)


earaghbidikashani (approved):

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
```diff
@@ -114,6 +114,19 @@ def test_all_eks_addons_gated_by_cluster_addons() -> None:
     )
 
 
+def test_karpenter_can_read_generated_instance_profiles() -> None:
+    """Karpenter's termination reconciler probes generated <cluster>_* profile
+    names with GetInstanceProfile; a 403 there blocks EC2NodeClass deletion (#349).
+    """
+    iam_tf = (TEMPLATE_PATH / "engine" / "iam.tf").read_text()
+    statement = re.search(r'sid\s*=\s*"AllowInstanceProfileGet"(.*?)(?=\n  statement|\n\})', iam_tf, re.DOTALL)
```
bhanderig: non-blocking:

This guard doesn't fail for the bug it's guarding. The lookahead's nearest boundary is the `\n  statement` of `AllowInstanceProfileList`, which sits *after* that statement's 4-line comment — so `group(1)` runs past the Get statement's closing `}` and swallows the neighbouring comment block. The actual capture on this branch ends with:

```
...cluster_name}_*",\n    ]\n  }\n\n  # Karpenter >=1.7 runs an instance-profile garbage-collection reconciler that calls\n  # iam:ListInstanceProfiles ...
```

Which means the assertion is satisfied by the ARN pattern appearing anywhere in that trailing comment, not just in `resources`. Reverting the grant while leaving a mention behind keeps it green:

```
ORIGINAL:               test passes = True
MUTANT (grant removed): test passes = True
```

Two smaller things while you're in here: it only asserts the wildcard, never that `aws_iam_instance_profile.karpenter_node.arn` is still in the list — dropping that would break the reconciler's `Get` on `Status.InstanceProfile` with the test still passing — and this file already has `_extract_resource_block` doing brace-matching precisely to avoid boundary regexes.

Suggest an `_extract_statement_block(content, sid)` alongside it that brace-matches the block, strips comment lines, and asserts **both** ARNs. If you'd rather keep it minimal, tightening the lookahead to `(?=\n  \})` plus the second assertion closes the false-negative.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
```diff
@@ -353,10 +353,17 @@ data "aws_iam_policy_document" "karpenter_controller" {
...
+  # finalizer, wedging the NodeClass in Terminating forever.
   statement {
-    sid       = "AllowInstanceProfileGet"
-    actions   = ["iam:GetInstanceProfile"]
-    resources = [aws_iam_instance_profile.karpenter_node.arn]
+    sid     = "AllowInstanceProfileGet"
+    actions = ["iam:GetInstanceProfile"]
+    resources = [
+      aws_iam_instance_profile.karpenter_node.arn,
+      "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:instance-profile/${module.eks_cluster.cluster_name}_*",
```
bhanderig, line 367: non-blocking:

This grant matches `LegacyInstanceProfileName` only — `fmt.Sprintf("%s_%d", clusterName, hash(region+nodeClass.Name))` at path `/`. Worth noting that v1.13.1 creates *new* managed profiles under an IAM path instead (`FormatPath("karpenter", region, clusterName, nodeClassUID)`), whose ARNs are `instance-profile/karpenter/<region>/<cluster>/<uid>/<name>` and are **not** matched by `<cluster>_*`. That's fine today — only `ListProfiles` touches those and it's granted on `*` — but it means correctness here rests on two upstream implementation details behind a bumpable pin: that the legacy probe still runs in pre-created-profile mode, and that its name is still the flat `<cluster>_<hash>`. Upstream has already moved new profiles off that scheme once.

What makes me want it written down rather than assumed: this is the second reactive grant of the same class. `git log -S'AllowInstanceProfileList'` lands on `b6f41a9` (#330 `feat: combined pool`) — the same commit that bumped `karpenter_version` 1.3.8 → 1.13.1, with the comment recording that the controller "floods AccessDenied" without it. So a 10-minor jump rode along inside a feature PR, needed a new IAM grant, and still missed this one. There's no `dependabot`/`renovate`, and `karpenter_version` appears nowhere in `AGENT.md`, `docs/`, `scripts/`, or the `justfile`, so nothing prompts a policy re-check on a bump.

Concretely: naming `EC2NodeClass.LegacyInstanceProfileName` and karpenter v1.13.1 in this comment gives a future bump something to grep — the `AllowInstanceProfileList` comment above already sets that precedent with "Karpenter >=1.7". Separately, is upgrade hygiene (diff the controller policy against upstream's reference policy when moving the pin) tracked in an issue anywhere? Not asking to widen this PR — just want it captured so the third one doesn't get found in production too.

bhanderig (dismissed):
LGTM with few nits

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
```diff
@@ -114,6 +114,19 @@ def test_all_eks_addons_gated_by_cluster_addons() -> None:
     )
 
 
+def test_karpenter_can_read_generated_instance_profiles() -> None:
```
JGuinegagne, line 134: non-blocking: UT is a fairly weak protection against regression on such a complex behavior. As @bhanderig explains below, a future Karpenter bump may introduce other AWS API calls that may require more permission.

Ideally, we would test the actual functional behavior of flipping on/off the `gpu_enabled` flag, and watch the node pool purge (as in a mutating e2e test).

Swapping nodepools smoothly is important enough a feature to warrant such an e2e test, even if it might be hard to write & flaky at first.

JGuinegagne (dismissed):

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
andrii-i: Thanks, good point, adopted the proposed `_extract_statement_block` version

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/iam.tf
andrii-i: Makes sense, good point, updated the comment as suggested. Also filed https://github.com/jupyter-infra/jupyter-deploy/issues/358 to track the bump-time policy check.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_eks_template.py
andrii-i: Agreed, tracked in #357.

JGuinegagne (approved):

bhanderig (approved):

earaghbidikashani (approved):

PR #352, opened 2026-08-27
fix: drop instance-cpu allowlist from accelerator pools

### LINKED ISSUE #351: Accelerator pools cannot provision multi-GPU instances due to a hardcoded vCPU allowlist in the karpenter-nodepools chart
## Problem

The karpenter-nodepools chart [pins every workspace pool to instance sizes with 2 to 32 vCPUs](https://github.com/jupyter-infra/jupyter-deploy/blob/1617d76046561e4ec0af340ef29dfa51fb75cbba/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml#L92-L94) (`karpenter.k8s.aws/instance-cpu In ["2","4","8","16","32"]`), which makes sense for CPU pools as it keeps node sizes in a predictable range for bin-packing.

The requirement applies to every pool, including GPU pools: nodes must have one of those five vCPU counts, which does not fit multi-GPU instances (p4d.24xlarge has 96 vCPUs, p5.48xlarge has 192). An instance size not on the list does not pass the filter and cannot be provisioned, so a pool whose families have no listed size never creates a node, and its workspaces are left Pending with only Karpenter's "no instance types satisfy requirements" event as a trace.

## Proposed solution

Omit the instance-cpu requirement for accelerator pools. With the requirement gone, the pool's instance families and the workspace's GPU request would determine the node size, and Karpenter provisions the cheapest instance that satisfies the pending pod, so small workloads still get small nodes and a p4d.24xlarge appears only when a workspace actually requests that much.

The alternative would be keeping some form of vCPU allowlist for accelerator pools, either extended with the missing sizes or made a per-pool setting. This is not practical because GPU families come with arbitrary vCPU counts that change with every new family, so any list has to chase AWS instance shapes and goes stale, while the instance family and the GPU request already determine the node size.

andrii-i:
Fixes #351.

Omits the instance-cpu requirement for accelerator pools. With the requirement gone, the pool's instance families and the workspace's GPU request determine the node size, and Karpenter provisions the cheapest instance that satisfies the pending pod, so small workloads still get small nodes.

### DIFF
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
@@ -89,10 +89,11 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
-        {{- /* Template comment (not rendered): accelerator families ship sizes
-             above this allowlist (p4d.24xlarge = 96 vCPUs, p5en.48xlarge = 192),
-             so on a GPU pool it can only exclude instances; the family list and
-             the gpu limit already pin the node size there. */}}
+        {{- /* On an accelerator pool the workspace's GPU request and the family
+             list determine the node choice; vCPU is noise there, and accelerator
+             families ship sizes far above this allowlist (p4d.24xlarge = 96
+             vCPUs), so the allowlist could only exclude instances. Per-node
+             size is deliberately unbounded on accelerator pools. */}}
         {{- if not .gpu }}
         - key: karpenter.k8s.aws/instance-cpu
           operator: In
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
@@ -89,11 +89,9 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
-        {{- /* On an accelerator pool the workspace's GPU request and the family
-             list determine the node choice; vCPU is noise there, and accelerator
-             families ship sizes far above this allowlist (p4d.24xlarge = 96
-             vCPUs), so the allowlist could only exclude instances. Per-node
-             size is deliberately unbounded on accelerator pools. */}}
+        {{- /* Accelerator families ship sizes above this allowlist (p4d.24xlarge =
+             96 vCPUs), so it would block their provisioning (#351). Per-node size
+             is deliberately unbounded on accelerator pools (#352 review). */}}
         {{- if not .gpu }}
         - key: karpenter.k8s.aws/instance-cpu
           operator: In
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
@@ -89,9 +89,15 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
+        {{- /* Template comment (not rendered): accelerator families ship sizes
+             above this allowlist (p4d.24xlarge = 96 vCPUs, p5en.48xlarge = 192),
+             so on a GPU pool it can only exclude instances; the family list and
+             the gpu limit already pin the node size there. */}}
+        {{- if not .gpu }}
         - key: karpenter.k8s.aws/instance-cpu
           operator: In
           values: ["2", "4", "8", "16", "32"]
+        {{- end }}
         - key: karpenter.sh/capacity-type
           operator: In
           values: ["on-demand"]
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
@@ -142,6 +142,15 @@ class TestKarpenterNodepoolsGpuRender(GoldenComparisonTestCase):
         self.assertEqual(self.gpu_pool["spec"]["limits"].get("nvidia.com/gpu"), "4")
         self.assertNotIn("nvidia.com/gpu", self.cpu_pool["spec"]["limits"])
 
+    def test_gpu_nodepool_no_instance_cpu_allowlist(self) -> None:
+        """Accelerator families ship sizes above the CPU allowlist (p4d.24xlarge
+        = 96 vCPUs); a GPU pool carrying it could never provision those families.
+        """
+        gpu_keys = {req["key"] for req in self.gpu_pool["spec"]["template"]["spec"]["requirements"]}
+        self.assertNotIn("karpenter.k8s.aws/instance-cpu", gpu_keys)
+        cpu_keys = {req["key"] for req in self.cpu_pool["spec"]["template"]["spec"]["requirements"]}
+        self.assertIn("karpenter.k8s.aws/instance-cpu", cpu_keys)
+
     def test_gpu_nodepool_instance_families(self) -> None:
         requirements = {req["key"]: req for req in self.gpu_pool["spec"]["template"]["spec"]["requirements"]}
         self.assertEqual(
--- a/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
+++ b/libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
@@ -163,6 +163,31 @@ class TestKarpenterNodepoolsGpuRender(GoldenComparisonTestCase):
 
 
 @_SKIP_WITHOUT_HELM
+class TestKarpenterNodepoolsAccelUnboundedRender(GoldenComparisonTestCase):
+    """An accelerator entry without maxGpus renders no instance-cpu allowlist and
+    no nvidia.com/gpu limit: per-node size is deliberately unbounded (#352 review).
+    """
+
+    rendered: ClassVar[str]
+    pool: ClassVar[dict[str, Any]]
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.rendered = _helm_template(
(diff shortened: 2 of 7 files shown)


libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -89,9 +89,15 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
+        {{- /* Template comment (not rendered): accelerator families ship sizes
+             above this allowlist (p4d.24xlarge = 96 vCPUs, p5en.48xlarge = 192),
+             so on a GPU pool it can only exclude instances; the family list and
+             the gpu limit already pin the node size there. */}}
```
bhanderig: nit (comment accuracy): `max_gpus` is optional, so "the gpu limit already pin[s] the node size" doesn't hold for every accelerator pool. `platform_karpenter.tf` emits `gpu = "true"` for every `accelerator = "nvidia"` entry but `maxGpus` only when the key is present, `variables.tf` enforces `max_gpus ⟹ accelerator` and never the reverse, and `autoscaling.md` documents the absence as intentional ("absent means unbounded, up to the account's service quota").

For that shape this leaves no `instance-cpu` allowlist *and* no `nvidia.com/gpu` limit — `instance_families` + `amd64` are the only node-shape constraints. On the documented `p4d,p5,p5en` example that reaches `p5.48xlarge` (192 vCPUs).

Suggest leaning only on what's always present:

```
     so on a GPU pool it can only exclude instances; the family list and
     the workspace's GPU request determine the node size there. */}}
```

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -89,9 +89,15 @@ spec:
...
+        {{- if not .gpu }}
         - key: karpenter.k8s.aws/instance-cpu
           operator: In
           values: ["2", "4", "8", "16", "32"]
```
bhanderig, line 98: Question, not a request. #351 rejects both alternatives because "any list has to chase AWS instance shapes and goes stale" — sound for an `In` list, but `instance-cpu` also takes `Gt`/`Lt` with a single value, which bounds without enumerating and can't go stale. Was that considered?

Asking because `max_cpu` can't stand in for a per-node bound: Karpenter's limits are checked against *current* consumption and are "eventually consistent, which can result in overrun." These pools sit at zero routinely (`consolidationPolicy: WhenEmpty`) — exactly where that check passes trivially. On the built-in pool (`max_cpu = "64"`, `g4dn,g5,g6,g6e,g7,g7e`), `g5.48xlarge`/`g6e.48xlarge` are 192 vCPUs, 3× the stated ceiling in one node. The allowlist made that structurally impossible.

Not reachable with the stock `jupyterlab-gpu` card (`gpus = "1"`, cpu/memory pinned), so no live regression — reachable via a custom `workspace_templates` entry.

A bound off the existing knob, no new variable:

```yaml
{{- else }}
        - key: karpenter.k8s.aws/instance-cpu
          operator: Lt
          values: ["{{ add1 (atoi .maxCpu) }}"]
```

Still admits every shape a pool is sized for (`g6e.12xlarge` = 48 under `max_cpu = "64"`, `p5.48xlarge` = 192 under `"384"`). If unbounded-per-node is deliberate, worth saying so in the comment so it doesn't get relitigated.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -89,9 +89,15 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
+        {{- /* Template comment (not rendered): accelerator families ship sizes
+             above this allowlist (p4d.24xlarge = 96 vCPUs, p5en.48xlarge = 192),
+             so on a GPU pool it can only exclude instances; the family list and
+             the gpu limit already pin the node size there. */}}
+        {{- if not .gpu }}
```
bhanderig, line 95: nit: `gpu` gains a second, unrelated effect and two written contracts go stale — `values.yaml` ("`gpu` (\"true\" adds the nvidia.com/gpu.present node label…)") and `platform_karpenter.tf` ("The chart's gpu value adds the nvidia.com/gpu.present node label; … untouched by the accelerator rename"). Both describe one purpose; it now also gates node sizing.

That second comment also records a deliberate `gpu` ≠ `accelerator` split, and this gate is accelerator-*shaped*, not nvidia-specific. Moot today since `variables.tf` pins `accelerator` to exactly `"nvidia"`. Stops being moot with a second device stack: `gpu` would be unset for a Neuron pool, `trn1.32xlarge` is 128 vCPUs, and it silently re-inherits #351.

Updating the two comments is enough for now.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
```diff
@@ -142,6 +142,15 @@ def test_gpu_nodepool_gpu_limit(self) -> None:
         self.assertEqual(self.gpu_pool["spec"]["limits"].get("nvidia.com/gpu"), "4")
         self.assertNotIn("nvidia.com/gpu", self.cpu_pool["spec"]["limits"])
 
+    def test_gpu_nodepool_no_instance_cpu_allowlist(self) -> None:
```
bhanderig, line 145: Assertion and docstring are the right shape, and the golden update keeps the `spec.template` hash-drift invariant covered.

Gap is in the fixtures: both `values_karpenter_cpu.yaml` and `values_karpenter_cpu_gpu.yaml` carry `maxGpus`, and the latter is pinned to "the GPU entry exactly as the `enable_default_gpu_pool` synthesis emits it" — so `accelerator: nvidia` with no `max_gpus` renders in no test. After this PR that's the shape with the fewest constraints of any pool the template can produce (no `instance-cpu`, no `nvidia.com/gpu` limit), and it's the config the template comment's rationale doesn't cover.

In scope here, or follow-up issue?

bhanderig (dismissed):
LGTM with few nits

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
JGuinegagne: I would say Gaurav that for accelerator-enabled nodes, vCPU shouldn't be the deciding factor. The GPU memory or number of GPUs will be the determining factors.

In that sense, removing vCPU limits for accelerator node provisioning decision is likely the right simplification.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
```diff
@@ -89,9 +89,15 @@ spec:
         - key: karpenter.k8s.aws/instance-family
           operator: In
           values: {{ .instanceFamilies | toJson }}
+        {{- /* Template comment (not rendered): accelerator families ship sizes
```
JGuinegagne: nit: remove the `Template comment (not rendered)` please;

Then add context as to why number GPU / GPU memory are the determining factor for GPU-node provisioning decisions, vCPU is just noise in that context.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
JGuinegagne: different PR, but: should we raise the max value? It seems fairly low, we may want to allow provisioning more powerful CPU instances for compute-intensive work (e.g. data manipulation type of work).

JGuinegagne (dismissed):
LGTM!

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
andrii-i: Thanks, done.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
andrii-i: @bhanderig Good question to consider. Per-node size on accelerator pools is deliberately unbounded and the template comment now records that: https://github.com/jupyter-infra/jupyter-deploy/commit/81d73e9e288566658b45a378d57428c4a2eef02a. A cap tied to max_cpu would also recreate #351 for any pool whose max_cpu sits below a family's only size (your 64 vs p4d's 96).

andrii-i: @JGuinegagne Agreed, filed #359.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
andrii-i: Done.

libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/charts/karpenter-nodepools/templates/nodepools.yaml
andrii-i: Both updated.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/unit/test_chart_render.py
andrii-i: In scope, added.

JGuinegagne (approved):

bhanderig (approved):

earaghbidikashani (approved):

PR #353, opened 2026-08-27
feat: jupyterlab template

JGuinegagne:
This PR adds the `tf-aws-ec2-jupyterlab` template: a single-user JupyterLab on a remote EC2 instance, reached from the user's laptop through the local client proxy (#347/#348) over pinned self-signed TLS and authenticated with the caller's AWS identity. **AWS credentials are the only prerequisite**: no domain, Route53, ACM, or OAuth app. There is no public URL and no Elastic IP: the proxy resolves the instance's public IP live and pins its certificate. The security group opens `:443` to all; the boundary is the pinned TLS plus a short-lived STS-identity token, not the network layer.

Depends on: #348

### User experience
- new rung-1 template (single user, AWS-creds-only):
    - `jd init . -E terraform -P aws -I ec2 -T jupyterlab` (will be just `jd init .` in the future)
    - `jd config`: works, no required variables
    - `jd up`
    - `jd open`
- `jd up` provisions the instance, generates a self-signed cert, and starts the docker services; no public URL, no EIP
- `jd open` starts the local proxy and opens `http://127.0.0.1:PORT/lab`; there is no browser sign-in
- access is authorized by **AWS IAM identity**: the `auth-sidecar` checks the caller's ARN against `auth_arn_allowlist` (defaults to the deployer's ARN); add ARNs to grant others
- the jupyter image ships in two package-manager flavors: `uv` (default) and `pixi`
- host/server lifecycle over SSM: `host start`/`stop`/`restart`/`connect`, `server logs`/`exec`/`connect`

### Implementation
1. new package `libs/jupyter-deploy-tf-aws-ec2-jupyterlab/` (`template.py`, `pyproject.toml`, exposing the `jupyter_deploy.terraform_templates` entry point) + root workspace wiring (`pyproject.toml`, `uv.lock`)
2. `template/engine/`: EC2 in the default VPC — `main`/`variables`/`outputs` plus `iam`/`cert`/`commands`/`services`/`waiter.tf`; no EIP/Route53 (proxy resolves the live IP and pins the cert)
3. `template/engine/modules/`: `ami_al2023`, `ec2_iam_role`, `ec2_instance`, `network`, `s3_bucket`, `volumes`
4. `services/auth-sidecar/` (Go ForwardAuth): replays the client-presigned `sts:GetCallerIdentity` token to the pinned STS host, enforces the `x-k8s-aws-id` binding, allowlists the returned ARN, and TTL-caches positive results; long term, we'll move it to its own repo and publish the image.
5. `services/docker-compose.yml.tftpl` + cloudinit: `jupyter` (:8888), `traefik` (TLS on :443), `auth-sidecar`, `fluent-bit`, `logrotator`; `generate-cert.sh.tftpl` mints the self-signed cert onto the EBS root volume of the instance and publishes the PEM to an SSM cert-pin parameter the proxy reads
6. `services/jupyter/` and `services/jupyter-pixi/`: the two package-manager image variants
7. `manifest.yaml`: `host.*`/`server.*` SSM commands, `proxy.connect-info` (resolve-endpoint + read cert pin + mint token), and `open: {mode: proxy, path: /lab}`; `variables.yaml`
8. `modules/network`: security group opens `:443` to `0.0.0.0/0` — pinned TLS + the STS token are the boundary
9. SSM-only host access: no `key_pair_name` variable; `AGENT.md`/`TROUBLESHOOT.md` document that only `host connect`/`server connect` need the Session Manager plugin
10. unit tests: `manifest.yaml`/`variables.yaml` parity, template packaging, and version (`tests/unit/`)

### Testing
- unit: `tests/unit/` — manifest/variables/template/version parity (`just unit-test`)
- manual: deployed end-to-end against a live instance — `jd config` (terraform plan), `jd up`, then `jd open` tunnels to `/lab` through the proxy and runs a notebook
- ran simple e2e tests (will commit in the next PR) that open the app via the proxy, and runs a simple notebook

### Screenshots
**jd config --help**
jd-config-help

**jd show (header)**
jd-show

### Screen recording
**jd open**
https://github.com/user-attachments/assets/630b2390-104f-42af-917d-2bd5407e4371

**jd open --detached**
https://github.com/user-attachments/assets/a4ed26d3-5254-4fd0-887e-cc0971d803c0


libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/generate-cert.sh.tftpl
```diff
@@ -0,0 +1,53 @@
...
+# data volume so a stop/start never rotates it, then publish ONLY the public PEM to the
+# cert-pin SSM parameter that `jd proxy connect-info` reads. The private key never leaves
+# the instance. The pin is on the cert, not the address (the client disables hostname
+# verification), so the SAN is irrelevant.
+set -e
+
+mkdir -p /var/log/jupyter-deploy
+exec > >(tee /var/log/jupyter-deploy/generate-cert.log) 2>&1
+
+CERT_DIR="/mnt/jupyter-data/.jupyter-deploy-tls"
```
bhanderig, line 12: `[Security]` The private key lands in the notebook's data volume, and a redeploy hands it to the notebook user.

`/mnt/jupyter-data` is bind-mounted as `/home/jovyan` (docker-compose.yml.tftpl:45) into a container running as `service-user` (:3). The SSM document runs `CloudInit` before `GenerateCertificates` (services.tf:303 → :341), and `CloudInit` does `chown -R service-user:service-user /mnt/jupyter-data` unconditionally (cloudinit.sh.tftpl:143).

First boot is fine — the chown precedes key creation. But the association is replaced on any `scripts_files_hash`/`ssm_embedded_hash`/`instance_type` change (services.tf:451-476), so on the next `jd up`: the chown takes `key.pem`, this script takes the `else` branch, and the `chmod 600` at line 29 never re-runs. Result: `key.pem` mode 600 owned by `service-user`, readable at `/home/jovyan/.jupyter-deploy-tls/key.pem` from any notebook cell.

That's the documented multi-user path, not a rare one — `arn_allowlist` feeds `docker-compose.yml.tftpl`, which is in `local.all_script_files` (iam.tf:16-19), so "add a teammate's ARN and `jd up`" triggers it.

Not an escalation for an allowlisted user, who already has host root via the Docker socket (see comment on docker-compose.yml.tftpl:46) — but a defence-in-depth regression that becomes the live path if that socket is ever tightened.

Preferred fix: move cert+key off the notebook volume (e.g. `/etc/jupyter-deploy/tls`). If the stop/start persistence described at line 2 is why they live there, the minimum fix is hoisting `chown root:root` + `chmod 600` out of the `if` so both re-assert on every run.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/jupyter/jupyter_server_config.py
```diff
@@ -0,0 +1,21 @@
...
+
+c.Application.log_level = "INFO"
+
+# This server is reached through the local jupyter-deploy client proxy: the browser's
+# Origin is the proxy's loopback (http://127.0.0.1:<random-port>) while the server sees
+# the instance IP as Host, so Jupyter's same-origin check blocks every websocket and
+# mutating API call. Allow the loopback origin on any port (the proxy picks a fresh port
+# each run). The real access boundary is the auth sidecar (STS-identity ForwardAuth) plus
+# pinned TLS, not this CORS check.
+c.ServerApp.allow_origin_pat = r"^https?://(127\.0\.0\.1|localhost)(:\d+)?$"
```
bhanderig: `[Security]` This accepts any page served from `http://127.0.0.1:<any-port>` or `http://localhost:<any-port>`.

It's needed today because the proxy forwards `Origin` verbatim (absent from `DROP_FROM_REQUEST_HEADERS`, constants.py:50-60), so Jupyter sees a loopback origin against an instance-IP `Host`. If the proxy instead rewrote `Origin`/`Referer` to `https://{bundle.host}:{bundle.port}`, the default same-origin check would pass and this line could be deleted. That's a `utils.py` change in #348, but this is the config that depends on it.

Why close it rather than accept it: Jupyter runs with no token auth (jupyter-start.sh:20, `--IdentityProvider.token=`), cookies ignore port so a page on another loopback port shares the jar, and Jupyter's `_xsrf` cookie is not `HttpOnly`. An unprivileged local process that binds a loopback port and gets one page loaded reaches kernel-execute.

Ceiling, in fairness: the proxy's loopback listener has no auth of its own and injects the credential, so any local process can already `curl http://127.0.0.1:<port>/api/contents` with no `Origin` at all. That's #348's design and it dominates this — the marginal exposure here is the browser leg.

The same line is at `services/jupyter-pixi/jupyter_server_config_pixi.py:12` and needs whatever lands here.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/fluent-bit/fluent-bit.conf
```diff
@@ -0,0 +1,45 @@
...
+    Format              plain
+
+[OUTPUT]
+    Name                file
+    Match               docker.traefik
+    Path                /logs
+    File                traefik.log
+    Format              plain
+
+[OUTPUT]
```
bhanderig, line 33: `[Observability]` Nothing matches `docker.auth-sidecar`, so the auth boundary's logs are dropped.

Compose tags are `docker.jupyter`, `docker.auth-sidecar` (:89), `docker.traefik`, `docker.log-rotator`. This file covers the first, third, fourth and `docker.oauth` — with no `Match *` fallback. So every `auth denied:` line from `main.go` is discarded: it never reaches `/var/log/services/`, which puts it outside the log-rotator's scope and outside `log_files_retention_days`.

Not a total blackout — `jd server logs auth-sidecar` still works (Docker's dual-logging cache has been on by default since 20.10, so `docker logs` reads the ring buffer despite the `fluentd` driver), and Traefik's `accessLog` records the 401/403. What's lost is which ARN was rejected and which of `verify()`'s failure modes fired, and that survives only in a ~10 MB ring buffer.

Suggest adding a `docker.auth-sidecar` output. Genuine question rather than a request: is `docker.oauth` intentional? There's no oauth2-proxy in this template, so it looks carried over from `tf-aws-ec2-base` — fine to leave if the two configs are deliberately kept diffable. A `Match *` fallback would also stop the next service from vanishing silently.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/main.tf
```diff
@@ -0,0 +1,126 @@
...
+  }
+  combined_tags = merge(
+    local.default_tags,
+    var.custom_tags,
+  )
+  doc_postfix = random_id.postfix.hex
+
+  # ARNs authorized to reach JupyterLab through the auth sidecar. Defaults to the
+  # deploying identity; extra ARNs can be added via var.auth_arn_allowlist.
+  auth_arn_allowlist = distinct(concat([data.aws_caller_identity.current.arn], var.auth_arn_allowlist))
```
bhanderig: `[Reliability]` The allowlist can't match the kind of ARN `variables.tf` tells users to add.

variables.tf:147-148 says "Add ARNs here to grant access to other IAM users or roles (e.g. a teammate's role)". A user adds `arn:aws:iam::123456789012:role/Dev`. But `GetCallerIdentity` returns `arn:aws:sts::123456789012:assumed-role/Dev/<session-name>`, and main.go:245-251 is a verbatim membership test with explicitly no prefix/wildcard matching. The entry is dead on arrival and the teammate gets a 403 with no diagnosable cause. EKS access entries, which use this same token construct, key on the IAM role ARN and normalize for exactly this reason.

Separately, the deployer's own entry is session-name dependent — `data.aws_caller_identity.current.arn` bakes it in at apply time. Stable under Identity Center and instance profiles; not under `aws-vault`, ad-hoc `assume-role` with a generated `--role-session-name`, or deploy-from-CI-then-`jd open`-from-laptop. And the recovery (`jd config && jd up`) is the redeploy that triggers the generate-cert.sh.tftpl issue.

Suggested fix, in the sidecar so it stays a property of the auth layer: derive `arn:aws:iam::<acct>:role/<Role>` from the assumed-role ARN and test both forms against the allowlist. Makes the documented entry work, makes the deployer's entry session-independent, and keeps the no-wildcard guarantee.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main.go
```diff
@@ -0,0 +1,287 @@
...
+	req.Header.Set(bindingHeader, binding)
+	req.Header.Set("Accept", "application/xml")
+
+	resp, err := s.http.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("STS replay failed: %w", err)
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(io.LimitReader(resp.Body, 1<<16))
+	if resp.StatusCode != http.StatusOK {
```
bhanderig, line 310: `[Reliability]` STS being unavailable is reported to the user as an authentication failure.

This line and the `s.http.Do` error above it collapse `429`, STS `5xx`, DNS failures and TCP resets into one error, and handleAuth:239-244 turns every error into `401`. There's no retry.

The cache holds only positive results and the client re-mints every ~45s (`DEFAULT_TOKEN_EXPIRY_SECONDS=60` − `DEFAULT_REFRESH_MARGIN_SECONDS=15`), so every fresh token is a cache miss that hard-depends on STS. A ~90s STS blip 401s every request in that window, the kernel websocket is torn down, and JupyterLab surfaces "session ended" rather than "retry" — a running notebook has a liveness dependency on STS at 60-second granularity with no grace.

Suggest: network errors and `429`/`5xx` → `503` with `Retry-After`, `401` reserved for genuine signature/binding/expiry rejections, plus one bounded retry on the transient class. It also makes the `auth denied:` log actionable, which today can't separate someone probing from STS throttling.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/docker-compose.yml.tftpl
```diff
@@ -0,0 +1,186 @@
...
+      - "traefik.http.routers.jupyter.entrypoints=websecure"
+      - "traefik.http.routers.jupyter.tls=true"
+      - "traefik.http.services.jupyter.loadbalancer.server.port=8888"
+      - "traefik.http.services.jupyter.loadbalancer.healthcheck.path=/lab/favicon.ico"
+      - "traefik.http.services.jupyter.loadbalancer.healthcheck.interval=30s"
+      - "traefik.http.services.jupyter.loadbalancer.healthcheck.timeout=10s"
+      - "traefik.http.routers.jupyter.middlewares=forwardauth@docker"
+    volumes:
+      - /mnt/jupyter-data:/home/jovyan
+      - /var/run/docker.sock:/var/run/docker.sock
```
bhanderig, line 51: `[Security]` The RW Docker socket makes `auth_arn_allowlist` a grant of root on the instance.

With `group_add: ${DOCKER_GID}` (:5-6) and a `root:docker 660` socket, any notebook cell can `docker run --privileged -v /:/host`. That nullifies `user:` (:3), `no-new-privileges:true` (:21-22) and cloudinit's `nproc`/`nofile` limits. Traefik and fluent-bit mount the same socket `:ro` (:132, :151); only Jupyter gets it writable.

Inherited verbatim from `tf-aws-ec2-base:46`, so not a regression. What changed is the trust model: variables.tf:145-148 presents adding an ARN as the way to share the notebook, and on this template that grant is root-on-EC2. On base the equivalent was mediated by GitHub OAuth into a single-tenant box.

Question, since I can't derive the intent: does JupyterLab need the socket here — docker-based kernels, `docker build` from a notebook — or did it come along with the compose file? If it's needed, variables.tf should state plainly that an allowlisted ARN gets host-level access. If not, dropping the mount and the `group_add` also closes the `key.pem` exposure noted on generate-cert.sh.tftpl.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/cert.tf
```diff
@@ -0,0 +1,23 @@
...
+# check_hostname disabled — the pin is on the cert, not the address).
+#
+# Terraform owns the parameter's identity (a stable name) but NOT its value: it seeds a
+# placeholder and ignores subsequent value changes, so the real PEM never lands in state
+# and no apply-time wait for cert generation is needed.
+resource "aws_ssm_parameter" "cert_pin" {
+  name        = "/jupyter-deploy/${local.doc_postfix}/cert-pin"
+  description = "Public PEM of the instance self-signed cert; written by the instance at boot."
+  type        = "String"
+  tier        = "Advanced" # PEMs can exceed the 4KB Standard-tier limit
```
bhanderig: `[Efficiency]` Advanced tier isn't needed, and the stated reason doesn't hold for this payload.

What's stored is one self-signed leaf, RSA-2048, minimal subject, one SAN extension (generate-cert.sh.tftpl:23-28) — roughly 1.2 KB of PEM against the Standard tier's 4,096-character limit. RSA-4096 would still be ~1.9 KB. There's no chain to publish either: the client loads this exact PEM as its sole trust anchor with `check_hostname = False` (tls.py:29-31).

Cost is $0.05/parameter/month — trivial per deployment, but a recurring line item on the template whose pitch is that AWS credentials are the only prerequisite.

Reason to fix now rather than file it: per AWS's docs, Advanced → Standard isn't a supported in-place change, so once deployments exist this needs a delete/recreate, during which the pin is absent and `jd open` fails until the instance republishes.

Also `--tier Advanced` at generate-cert.sh.tftpl:41 has to stay in sync with this, and `ignore_changes` covers only `value` — if they diverge, `put-parameter` flips the tier and Terraform reports perpetual drift. If there is a plan to exceed 4 KB, keep Advanced and just fix the comment's arithmetic.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# Build stage
+FROM golang:1.26 AS builder
+ARG TARGETOS
+ARG TARGETARCH
+
+# Bypass Go proxy due to corporate network issues
+ENV GOPROXY=direct
```
bhanderig: `[Review-risk]` This builds on the customer's EC2 instance, where "corporate network issues" don't apply — the comment describes a constraint from another environment and will read oddly in a published template.

It's also a no-op with a cost: the module is stdlib-only with no `go.sum`, so `go mod download` (line 12) fetches nothing, and `GOPROXY=direct` forfeits the module proxy and cache for no gain. Suggest dropping both the `ENV` and the comment.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# instance builds this via docker-compose), matching the container it ships in.
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -trimpath -o auth-sidecar .
+
+# Use distroless as minimal base image to package the binary
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY --from=builder /workspace/auth-sidecar /auth-sidecar
+USER 65532:65532
+
+# ForwardAuth + server-observed-IP endpoint
```
bhanderig: `[Review-risk]` Stale comment — main.go:281-282 registers only `/auth` and `/healthz`; there is no server-observed-IP endpoint (presumably removed when `aws.ec2.resolve-endpoint` took over live IP resolution). Suggest `# ForwardAuth endpoint`.

AGENT.md
```diff
@@ -124,6 +124,19 @@ Always run from the root of the repository:
 2. Run unit tests: `just unit-test`
    - Runs `uv run pytest`
 
+## Running the client-proxy functional tests
```
bhanderig, line 164: `[Review-risk]` AGENT.md carries a section per package — "Base template package", "EKS OIDC template package", "CI infrastructure template package", "E2E Pytest plugin package" — but the new `tf-aws-ec2-jupyterlab` package gets none, even though this PR edits the file (for the client-proxy test track). Without one, the next contributor has no statement of its layout rules: `variables.tf`-only variable blocks, the `services/` split, and the `/opt` vs `/home/jovyan` build-time gotcha, which applies here too.

Same theme, and happy to drop all of it if there's a tracking issue to link — this is the only template missing them:
- `tests/e2e/` (both other templates have it; the PR body says next PR)
- the `e2e-jupyterlab*` workflow set and `release-jupyterlab.yml` (the others have all five plus a release workflow, so this package can't be published yet)
- `docs/source/templates/aws-ec2-jupyterlab-template/`, plus entries in `templates/index.md` and `concepts/templates.md` — the docs in this diff are all #348's proxy pages

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/README.md
```diff
@@ -0,0 +1,38 @@
...
+- **Data path:** the browser talks to a local proxy over `http://localhost`; the proxy
+  talks to the instance's Traefik on `:443` over pinned self-signed TLS (the pin is on the
+  cert, not the address, so a new public IP after a stop/start is a non-event).
+- **Cert pin:** the instance generates a long-lived self-signed cert at boot (private key
+  persisted on the EBS data volume) and publishes only the public PEM to an SSM parameter
+  that `jd proxy connect-info` reads live.
+- **Auth:** `jd proxy connect-info` mints a `k8s-aws-v1` STS-identity token; a ForwardAuth
+  sidecar behind Traefik validates it (STS replay + ARN allowlist + `x-k8s-aws-id` binding).
+  No shared secret is stored anywhere.
+- **Network:** the security group allows `:443` only, reconciled to the caller's `/32` on
```
andrii-i: This describes the per-caller security-group reconcile that was dropped in #348.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/AGENT.md.template
```diff
@@ -0,0 +1,167 @@
...
+### Host execution commands (when host is running)
+- `jd host exec -- CMD` - Execute commands on the instance
+- `jd host exec -- docker container list` - View the docker services
+- `jd host exec -- tail -50 /var/log/jupyter-deploy/LOG-FILENAME` - View command logs
+- `jd host exec -- tail -50 /var/log/services/LOG-FILENAME` - View historic service logs
+
+### Server execution commands (when host and target docker service are running)
+- `jd server logs` - View docker logs of the jupyter service
+- `jd server logs -n traefik -- -n 50` - View last 50 lines of the docker logs of the traefik service
+- `jd server logs -n auth-sidecar` - View docker logs of the auth-sidecar service
```
andrii-i: `jd server logs` has no `-n` option (the service flag is `-s`/`--service`), so both examples fail with `No such option: -n`. It should be `-s` for the service, as in `jd server logs -s traefik -- -n 50`.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/outputs.tf
```diff
@@ -0,0 +1,97 @@
...
+  value       = module.ec2_instance.ami
+}
+
+# Cert-pin transport: the SSM parameter NAME the proxy reads live (not the PEM itself).
+output "cert_pin_ssm_parameter_name" {
+  description = "Name of the SSM parameter holding the instance self-signed cert PEM."
+  value       = aws_ssm_parameter.cert_pin.name
+}
+
+# ARNs authorized to reach JupyterLab through the auth sidecar.
```
andrii-i: Update the header to say IAM role/user names?

andrii-i (dismissed):
Thanks @JGuinegagne, this looks good. A couple of non-blocking comments below.

JGuinegagne:
> ## roborev: Combined Review (`f1842ff`)
> ## Code Review Summary: Sound auth design; one Medium gap — the STS-identity auth boundary ships untested.
> 
> The reviewers examined the new EC2 JupyterLab template (STS-identity ForwardAuth over pinned self-signed TLS) and the hardened client proxy. The security review found the auth model largely sound: the sidecar replays the caller's presigned `sts:GetCallerIdentity` request to the pinned STS host and authorizes the ARN STS returns (account-scoped, exact-name allowlist, no wildcards), and the proxy origin gate correctly blocks the cross-origin mutation/WebSocket RCE path. The findings below the Medium bar (defense-in-depth notes on signed-header verification, docker.sock mount, origin-gate GET gap, unpinned boot downloads, README permission inaccuracy) are omitted per severity threshold.
> ### 🟡 Medium
> 
> **Auth-sidecar ships with zero tests and no CI target** `libs/jupyter-deploy-tf-aws-ec2-jupyterlab/.../services/auth-sidecar/main.go` (whole file); `AGENT.md` ("no CI target yet")
> 
> The auth-sidecar is the _entire_ access-control boundary for this template (STS-identity ForwardAuth + ARN/account allowlist), yet it has no automated tests and no CI target. The logic it encodes is exactly what regresses silently and dangerously: `parsePrincipal` ARN parsing (assumed-role vs user vs root/federated), account scoping in `allows`, the SSRF host-pin check, `X-Amz-Expires` bounds, the binding-header equality, and the positive-only cache re-check. A future edit that, e.g., loosens `parsePrincipal` to return a non-empty `kind` for an unexpected ARN shape would open the deployment with no test to catch it.
> 
> **Fix:** Add a Go test suite (`main_test.go`) covering `parsePrincipal`/`allows` (role, user-with-path, root, federated, cross-account, empty allowlist), the token pre-checks in `verify` (bad prefix, non-base64, wrong host, wrong action, out-of-bounds expiry, binding mismatch), and cache TTL/eviction. Wire a `go test` step into CI.
> 

Ack - long term (horizon 11/2026), we will move the STS auth-middleware to its own GitHub repository and add unit tests, integration tests etc. Short term, we will add indirect coverage via template E2E tests.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/main.tf
JGuinegagne: Thanks! Updated the variables / outputs to expose role & user names instead of full ARNs. This change makes it way more user-friendly, and also enforces same account IAM principal, which is safer given the single-user use case of this template.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/docker-compose.yml.tftpl
JGuinegagne: This is by design; it's a useful feature for users to have write access to the docker daemon, even if it clearly is an avenue for escalation of privilege. We accept this risk in order to support the feature.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/docker-compose.yml.tftpl
```diff
@@ -0,0 +1,209 @@
...
+      # bind the token to this deployment, scope to this account, allowlist IAM role/user
+      # names (unique per account regardless of path), and pin the STS endpoint.
+      - DEPLOYMENT_ID=${deployment_id}
+      - AWS_ACCOUNT_ID=${aws_account_id}
+      - ROLE_NAME_ALLOWLIST=${role_allowlist}
+      - USER_NAME_ALLOWLIST=${user_allowlist}
+      - AWS_REGION=${aws_region}
+      - STS_ENDPOINT=https://sts.${aws_region}.amazonaws.com
+    labels:
+      - "traefik.enable=true"
```
andrii-i, line 121: Traefik publishes containers with `traefik.enable=true` and a `loadbalancer.server.port` label and no router label by applying its [`defaultRule`](https://doc.traefik.io/traefik/reference/install-configuration/providers/docker/#defaultrule), a route matching the `Host` header `auth-sidecar-docker`, on every port it listens on and without TLS. So the sidecar is reachable from the internet over plain HTTP on `:443`. To reproduce with `traefik:latest` and these exact labels: `curl -H 'Host: auth-sidecar-docker' http://<instance-ip>:443/healthz` returns 200 from the sidecar, and 404 without the header. Only status codes are exposed, but still we should take the sidecar out of Traefik's routing: the `forwardauth` middleware calls it directly over the compose network (`http://auth-sidecar:4181/auth`), so it never needs a route, and these labels exist only because the middleware is declared through them. Defining the middleware in `traefik-dynamic.yml` (which Traefik already loads), pointing the jupyter router at `forwardauth@file`, and dropping this `labels:` block keeps the auth check exactly as it is and leaves nothing for Traefik to expose.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/docker-compose.yml.tftpl
```diff
@@ -0,0 +1,209 @@
...
+      start_period: 5s
+    networks:
+      - web
+    environment:
+      # The auth sidecar treats these as its whole interface (clean for later extraction):
+      # bind the token to this deployment, scope to this account, allowlist IAM role/user
+      # names (unique per account regardless of path), and pin the STS endpoint.
+      - DEPLOYMENT_ID=${deployment_id}
+      - AWS_ACCOUNT_ID=${aws_account_id}
+      - ROLE_NAME_ALLOWLIST=${role_allowlist}
```
andrii-i, line 116: The allowlist reaches the sidecar through this compose file, and the compose file is part of [`scripts_files_hash`](https://github.com/jupyter-infra/jupyter-deploy/blob/f1842ffb1118bf8f36591d1caac092a52a5a861f/libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/services.tf#L264). So changing it in any way forces a `docker compose up -d --force-recreate`, which restarts JupyterLab. We should give the allowlist a runtime path that recreates only `auth-sidecar`, as `jd users` does for `oauth` in the base template. And based on the comment at [`main.go` line 344](https://github.com/jupyter-infra/jupyter-deploy/blob/f1842ffb1118bf8f36591d1caac092a52a5a861f/libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main.go#L344), not redeploying for an allowlist change looks like the intended design.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/waiter.tf
```diff
@@ -0,0 +1,48 @@
...
+  await_server_indented = join("\n${local.await_indent_str}", compact(split("\n", local.await_server_file)))
+}
+
+resource "null_resource" "wait_for_instance_ready" {
+  triggers = {
+    # Instance parameters:
+    instance_id = module.ec2_instance.id
+    # the instance ID might be preserved even on VM swap
+    # add instance public IP (dynamic — no EIP; a new IP after stop/start is fine).
+    instance_ip    = module.ec2_instance.public_ip
```
andrii-i, line 19: Without an Elastic IP, AWS [releases the public IP when the instance stops and assigns a new one when it starts](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html), so this trigger changes on every stop/start and is empty while the instance is stopped. The comment above says a new IP after stop/start is fine, but a changed trigger replaces `null_resource.wait_for_instance_ready`, so after `jd host stop` a `jd up` with no other change re-runs the await script, which starts the instance and waits on SSM and status polling for several minutes. We should remove the `instance_ip` entry from this `triggers` map: the await script never uses the IP (its inputs are the instance id, association id, document name and region), and the remaining triggers already cover a replaced instance.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/main.tf
```diff
@@ -0,0 +1,134 @@
...
+  )
+  doc_postfix = random_id.postfix.hex
+
+  # Principals authorized to reach JupyterLab through the auth sidecar, by IAM name (unique
+  # per account regardless of path). The caller is merged into the matching list so switching
+  # callers produces no state diff (as long as every caller is declared in the right list).
+  # For an assumed-role caller, arn is arn:aws:sts::<acct>:assumed-role/<RoleName>/<session>
+  # (element 1 == RoleName); for a user, arn is arn:aws:iam::<acct>:user/<path...>/<name>.
+  caller_is_user   = can(regex(":user/", data.aws_caller_identity.current.arn))
+  caller_role_name = !local.caller_is_user ? element(split("/", data.aws_caller_identity.current.arn), 1) : ""
```
andrii-i, line 39: This is only designed to work for callers that are IAM roles or users (`:assumed-role/` or `:user/` ARNs), but nothing enforces that. Root and federated users are the right things to exclude, so we should add a `precondition` that fails the plan with a clear error for other caller types, and update the [`variables.tf` description](https://github.com/jupyter-infra/jupyter-deploy/blob/f1842ffb1118bf8f36591d1caac092a52a5a861f/libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf#L147) that currently says the deploying identity is always authorized.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/ec2_instance/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  # 2. Add buffer: max(33% of AMI size, AMI size + 10GB) - ensures at least 10GB headroom
+  # 3. Ensure we meet the minimum requirement (if specified)
+  # Result: Volume sizes scale naturally with AMI needs while maintaining adequate headroom
+  ami_root_size_gb      = try(local.root_block_device.ebs.volume_size, 5)
+  root_size_with_buffer = max(ceil(local.ami_root_size_gb * 1.33), local.ami_root_size_gb + 10)
+  root_volume_size_gb   = var.min_root_volume_size_gb != null ? max(var.min_root_volume_size_gb, local.root_size_with_buffer) : local.root_size_with_buffer
+}
+
+# EC2 instance
+resource "aws_instance" "ec2_jupyter_server" {
```
andrii-i, line 28: This instance gets a public IPv4 only because the default subnet auto-assigns one (the base template attached an Elastic IP instead). The template needs that address, both for the instance to reach SSM, S3 and STS and for the proxy to dial it, so we should request it explicitly rather than depend on a subnet setting that security baselines commonly turn off: `associate_public_ip_address = true` on the instance [overrides the subnet's attribute](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html).

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/manifest.yaml
```diff
@@ -0,0 +1,318 @@
...
+        label: Configuring EC2 instance
+        weight: 30
+        phases:
+        - enter-pattern: Waiting for SSM agent
+          label: Waiting for SSM agent
+          weight: 10
+        - enter-pattern: Cloudinit script is running
+          label: Configuring EC2 instance
+          weight: 35
+        - enter-pattern: Generating self-signed cert
```
andrii-i, line 313: These `enter-pattern`s are strings the CLI looks for in Terraform's output to advance the progress bar, and it [only looks for the next sub-phase in order](https://github.com/jupyter-infra/jupyter-deploy/blob/68dc9d36578576c7f21d36f0684e96e0de32ea31/libs/jupyter-deploy/jupyter_deploy/engine/supervised_phase.py#L149-L155). `Generating self-signed cert` is printed on the instance by `generate-cert.sh`, so it never appears in Terraform's output; the CLI keeps waiting for it and never reaches `Checking instance status`, whose line does appear. The deploy itself is fine, but the bar sits on `Configuring EC2 instance` through the whole docker build wait. We should drop this sub-phase and fold its weight into `Checking instance status`.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main.go
```diff
@@ -0,0 +1,415 @@
...
+// names are only unique within an account, so matching by name alone would authorize a same-named
+// principal in a different account.
+func (c *config) allows(arn string) bool {
+	account, kind, name := parsePrincipal(arn)
+	if account == "" || account != c.accountID {
+		return false
+	}
+	switch kind {
+	case "role":
+		return c.roleNames[name]
```
andrii-i, line 164: This lookup is case-sensitive, but IAM role and user names [are unique within an account without regard to case](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html), so `datascience` in the allowlist for a role created as `DataScience` is a silent, permanent 403. We should match names the same way, case-insensitively.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/commands.tf
```diff
@@ -0,0 +1,172 @@
...
+description: Execute a command inside a service container.
+parameters:
+  service:
+    type: String
+    description: "The service in which to execute the command (jupyter, traefik or auth-sidecar)."
+    default: jupyter
+    allowedValues:
+      - jupyter
+      - traefik
+      - auth-sidecar
```
andrii-i, line 97: The sidecar's image is [distroless](https://github.com/jupyter-infra/jupyter-deploy/blob/f1842ffb1118bf8f36591d1caac092a52a5a861f/libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/Dockerfile#L22), with no shell or coreutils, so `jd server exec -s auth-sidecar -- <cmd>` always fails with `executable file not found`, which says nothing about why. We should drop `auth-sidecar` from this list so the CLI rejects it with a clear message. Users can still inspect the sidecar with commands that don't run inside the container: `jd server logs -s auth-sidecar` for its output and `jd host exec -- docker inspect auth-sidecar` for its environment.

andrii-i (approved):
Thanks @JGuinegagne, re-approving. A few more nonblocking comments below, fine as follow-ups.

bhanderig (approved):
LGTM

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/docker-compose.yml.tftpl
JGuinegagne: woaw okay, will track this in subsequent PR but before the first release.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/main.tf
JGuinegagne: Good point, will enforce this rule before the next release. We can use [this mechanism](https://github.com/jupyter-infra/jupyter-deploy/blob/90510516b5b888201a8cca1505080aff07c72505/libs/jupyter-deploy-tf-aws-eks-oidc/jupyter_deploy_tf_aws_eks_oidc/template/engine/main.tf#L159) in `eks-oidc` template (fails at plan time if role or user not found).

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/manifest.yaml
JGuinegagne: yeah, I'll remove this phase. It would be instant for this template.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/commands.tf
JGuinegagne: Nice catch. Interestingly, `oauth2-proxy` image is also distroless, and we test exec w/ `oauth2-proxy --version` [here](https://github.com/jupyter-infra/jupyter-deploy/blob/90510516b5b888201a8cca1505080aff07c72505/libs/jupyter-deploy-tf-aws-ec2-base/tests/e2e/test_server.py#L156). However, there doesn't seem to be an equivalent w/ `auth-sidecar`.

PR #364, opened 2026-09-02
chore: jupyterlab e2e

JGuinegagne:
This PR adds simple e2e tests for the `tf-aws-ec2-jupyterlab` templates, along with plugin utilities to support them.

### Developer experience
- create a new project & run all the tests: `just test-e2e-jupyterlab`
- run tests against an existing deployment in local dir `sandbox-lab`: `just test-e2e-jupyterlab sandbox-lab`

### Implementation
1. `<plugin>/local_proxy`: add a `LocalProxyApplication` class w/
    1.1. `start()` method to start the local proxy
    1.2. `stop()` method to stop the local proxy
    1.3. `verify_jupyterlab_accessible()` method to assert that jupyterlab home page displays
2. `<plugin>/cli`: add proxy handling utilities
3. `<plugin>/client_proxy_app` add pytest fixture that starts the proxy, yields`LocalProxyApplication`, stops it afterwards
4. `<lab-template>/tests/e2e`:
    4.1. basic `conftest` for fixture definition
    4.2. `test_configuration` to ensure the terraform project is valid (+store defined, AGENT.md parses)
    4.3. `test_application` to open the app, open a basic notebook, run it

### Testing
- ran full e2e test suite against fresh app
- ran full e2e test suite against existing app


libs/pytest-jupyter-deploy/pytest_jupyter_deploy/cli.py
```diff
@@ -328,6 +329,64 @@ def get_jupyterlab_url(self) -> str:
...
+        """
+        result = self.run_command(["jupyter-deploy", "proxy", "status"])
+        for line in result.stdout.splitlines():
+            if "Proxy status:" in line:
+                status = line.split(":", 1)[1].strip()
+                return re.sub(r"\x1b\[[0-9;]*m", "", status)
+        raise ValueError("Could not parse proxy status from command output")
+
+    def stop_proxy(self) -> None:
+        """Stop any local proxy running for this project (no-op if none)."""
```
bhanderig: `[Reliability]` Not a no-op — `ProxyManager.stop()` raises `NoProxyFoundError` when nothing is running, and that's a handled CLI error, so `jd proxy stop` exits non-zero and `run_command` (`check=True`) turns it into `JDCliError`. This works today only because the fixture wraps both call sites in `contextlib.suppress(Exception)` (`plugin.py:658`, `:664`) — the no-op lives in the caller, not here. Same wording on `LocalProxyApplication.stop()` (`local_proxy/application.py:60`).

```suggestion
        """Stop the local proxy for this project.

        Raises:
            JDCliError: If no proxy is running. Callers using this for idempotent
                setup/teardown must suppress it — see the ``client_proxy_app`` fixture.
        """
```

justfile
```diff
@@ -485,6 +485,15 @@ e2e-all project_dir test_filter="" options="" no_cache="false" template=default-
 test-e2e-base project_dir="sandbox-e2e" test_filter="" options="":
     @just test-e2e {{project_dir}} "{{test_filter}}" "{{options}}" tf-aws-ec2-base
 
+# Run E2E tests for the JupyterLab template (tf-aws-ec2-jupyterlab)
+# Needs AWS credentials but NO test env vars and NO ci-dir (the local client proxy injects
+# the identity token itself — there is no browser sign-in).
+# Example: just test-e2e-jupyterlab sandbox-lab test_project_is_configurable   # config only
+# Example: just test-e2e-jupyterlab sandbox-lab test_configuration              # all config tests
+# Example: just test-e2e-jupyterlab sandbox-lab test_application                 # proxy + notebook in browser
+test-e2e-jupyterlab project_dir="sandbox-e2e" test_filter="" options="":
```
bhanderig, line 494: `[Reliability]` `sandbox-e2e` is the base/EKS sandbox dir per AGENT.md, and `test-e2e` only deploys fresh when the dir is absent **or empty**. If one already exists, a bare `just test-e2e-jupyterlab` silently enters existing-project mode and runs this suite against a base/EKS deployment — `jd proxy connect-info` isn't a base manifest command. The three examples above already use `sandbox-lab`.

```suggestion
test-e2e-jupyterlab project_dir="sandbox-lab" test_filter="" options="":
```

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_configuration.py
```diff
@@ -0,0 +1,131 @@
...
+
+@pytest.mark.cli
+def test_gitignore_generated_after_init(e2e_deployment: EndToEndDeployment) -> None:
+    """`.gitignore` is generated after `jd init` with the JD + terraform patterns."""
+    with undeployed_project(e2e_deployment.suite_config) as (project_path, cli):
+        gitignore_path = project_path / ".gitignore"
+        assert gitignore_path.exists(), f".gitignore should exist after init: {gitignore_path}"
+
+        content = gitignore_path.read_text()
+        assert ".jd-history/" in content
```
bhanderig, line 62: `[Observability]` The template writes seven patterns; this asserts six — `.jd-proxy/` is missing. It was added in #348 because the S3 store backup excludes by `.gitignore`, so without it `jd up`/`jd down` uploads every proxy runtime dir with its logs. This template creates one on every `client_proxy_app` test, so a regression would surface here first and nothing would catch it.

```suggestion
        assert ".jd-history/" in content
        # #348: without this, `jd up` backs up proxy runtime dirs (logs included) to S3.
        assert ".jd-proxy/" in content
```

(base's test has the same gap — separate backfill)

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_application.py
```diff
@@ -0,0 +1,36 @@
...
+from pytest_jupyter_deploy.notebook import delete_notebook, run_notebook_in_jupyterlab, upload_notebook
+
+
+def test_application_simple_python(
+    e2e_deployment: EndToEndDeployment,
+    client_proxy_app: LocalProxyApplication,
+) -> None:
+    """Run a simple Python notebook through the proxy-tunneled JupyterLab."""
+    # The client_proxy_app fixture already ensured the server is running and started the proxy.
+    client_proxy_app.verify_jupyterlab_accessible()
```
bhanderig, line 22: `[Security]` The auth-sidecar + short-lived STS token is this template's only perimeter, and nothing in the suite tests the negative path: a no-token request and an expired/malformed-token request should both be refused. Base pairs its happy path with `test_users.py` / `test_org_and_teams.py`, EKS with `test_routing_network_security.py`. As written, a sidecar regression — or the `allow_origin_pat` loopback wildcard from #353 — passes green. Fine as follow-up, can we track it?

justfile
JGuinegagne: disagree w/ that one, I would rather keep `sandbox-e2e` as the default dir for all templates

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_application.py
JGuinegagne: ack - will add -ve access tests in subsequent PRs

bhanderig (approved):
lgtm

PR #365, opened 2026-09-02
chore: proxy ci

JGuinegagne:
This PR stands up CI for the standalone client proxy and unbreaks the e2e image builds. a/ fixes the shared + CLI e2e image builds, which started failing once `jupyter-deploy-client-proxy` and the jupyterlab template joined the `uv` workspace; b/ adds proxy CI: b.1/ functional tests on every PR, b.2/ a `release-proxy` workflow, and b.3/ a CLI-with-proxy install smoke — and cuts the proxy to `0.1.0rc1`.

### Developer experience
- the client-proxy functional suite now runs on every PR (previously local-only via `just functional-test`)
- `just update-version proxy <patch|minor|major|version>` bumps the client-proxy package
- `release-proxy` workflow (`workflow_dispatch`) publishes the proxy to test-PyPI, then to PyPI

### Implementation
1. `.github/e2e-shared/Dockerfile` + `.github/e2e-cli/Dockerfile`: copy the `jupyter-deploy-client-proxy` and `jupyter-deploy-tf-aws-ec2-jupyterlab` pyproject (+ LICENSE/README stubs) in the dep layer, so `uv sync --all-packages` resolves them as workspace members
2. `.github/workflows/functional-test.yml`: reusable (`workflow_call`) job running `just functional-test` on the 3.12/3.13 matrix
3. `.github/workflows/ci.yml`: call it on every push/PR — the functional suite is excluded from the per-dir `test` run by root `norecursedirs`
4. `.github/workflows/release-proxy.yml`: build → lint → test → functional gate → Test PyPI → install-check → PyPI → GH release; `prerelease` auto-flagged for rc/a/b tags
5. `scripts/update_version.py`: add a `proxy` target (single-file bump)
6. `libs/jupyter-deploy-client-proxy/pyproject.toml`: `0.1.0` → `0.1.0rc1` (+ `uv.lock`)
7. update `.github/AGENT.md`
8. add GitHub pill for new package to root `README.md`

### Testing
- ci: client-proxy functional suite (21) + proxy unit (109) green
- manual: reproduced the `not a workspace member` failure, then confirmed `uv sync --all-packages` passes with the fixed dep-layer COPY set; `uv build` produces the `0.1.0rc1` sdist + wheel; new/changed workflow YAML validated

### DIFF
--- a/.github/workflows/release-proxy.yml
+++ b/.github/workflows/release-proxy.yml
@@ -1,0 +1,224 @@
+name: Release proxy (jupyter-deploy-client-proxy)
+
+run-name: Release jupyter-deploy-client-proxy
+
+# Mirrors release-plugin.yml (no live-deploy E2E gate — the proxy is cloud-blind). Its gate is
+# the offline functional suite; the Test PyPI install-check doubles as a CLI-with-proxy smoke.
+on:
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+      pkg-name: ${{ steps.get-version.outputs.pkg-name }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install UV and Python
+        uses: ./.github/actions/install-uv
+        with:
+          python-version: '3.13'
+          working-directory: '.'
+
+      - name: Build package
+        run: |
+          cd libs/jupyter-deploy-client-proxy
+          uv build
+
+      - name: Get package version
+        id: get-version
+        run: |
+          cd libs/jupyter-deploy-client-proxy
+          echo version=$(uv version --short) >> $GITHUB_OUTPUT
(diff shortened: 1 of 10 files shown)


michaelnchin (approved):
LGTM!

.github/workflows/release-proxy.yml
```diff
@@ -0,0 +1,203 @@
...
+        uses: astral-sh/setup-uv@v3
+
+      - name: Publish to Test PyPI
+        run: |
+          uv publish \
+            --publish-url https://test.pypi.org/legacy/ \
+            --check-url https://test.pypi.org/simple \
+            --trusted-publishing always
+
+  test-pypi-install:
```
andrii-i, line 97: With no `permissions:` block, this job inherits the workflow's `contents: write` and `id-token: write`. It only needs `permissions: contents: read`, since it only installs.

andrii-i (approved):
Thanks @JGuinegagne, this looks good. One nonblocking comment below.

.github/workflows/release-proxy.yml
JGuinegagne: Thanks Andrii. I'm following precedents from other workflows, but generally agree that we should tighten all CI workflows.

PR #366, opened 2026-09-03
test: gpu e2e in normal runs and e2e coverage expansion

### LINKED ISSUE #357: Enable GPU workspace tests in normal e2e runs
## Problem

The GPU workspace e2e tests from #341 only run when the `JD_E2E_GPU_ENABLED` environment variable is set. The e2e suite runs in three places today: the weekly scheduled workflow (`e2e-eks-oidc-canary.yml`), the e2e check inside the release workflows, and manually triggered runs. None of them set that variable, so the GPU tests are always skipped, and GPU changes ship with no coverage on real hardware: #349 and #351 both reached a release this way.

## Proposed solution

Run the GPU tests inside these existing runs. Nothing about when e2e runs changes: no new workflows, no runs on push.

Use the approach the base template already uses in `test_gpu.py`: the test itself changes the deployed project's configuration and restores it afterwards. Concretely, the test turns `enable_default_gpu_pool` on with `jd config` and `jd up`, runs the GPU workspace lifecycle test, then turns the flag back off and verifies the GPU node pool drains and its EC2NodeClass is deleted. Exercising the flag turning on and off is the point: creating and tearing down the pool are where regressions live (#349 was a teardown bug that a permanently enabled pool would never hit), and the teardown check is the e2e regression test asked for in the #350 review. Restoring the flag also leaves the shared test project as the test found it.

Costs stay roughly where they are. These runs already have a deployed project, so no additional Let's Encrypt certificate is used (certificates are only issued when a new project is deployed, and Let's Encrypt allows 5 per week for the same domain). The added cost per run is one GPU node for about 10 minutes (about $0.15 of g4dn on-demand) and two `jd config`/`jd up` cycles, about 20 to 30 minutes of extra run time.

The alternative would be a separate workflow dedicated to GPU testing that deploys its own project each run. That spends one of the 5 weekly certificates every time it runs and duplicates a deployment the existing runs already have, for the same coverage.

andrii-i:
Fixes #357

This PR turns on GPU coverage in the regular e2e runs: one mutating test module covers the GPU critical path end to end on both surfaces, kubectl and the web UI, each ending in a notebook run that proves torch detects CUDA, plus the pool lifecycle around them. It also adds a KEDA autoscaling test covering both scale-up and scale-down, and moves the generic test helpers into the pytest plugin.

All of this runs against a real EKS cluster created with jupyter-deploy, the same deployment the existing e2e workflows test against; GitHub Actions assumes an AWS role and drives the tests with pytest and a real browser. E2e runs in three places, unchanged by this PR: the weekly canary, the e2e gate inside the release workflows, and manual dispatches (including against a specific PR via the `ref` input); nothing runs on push or on pull requests. GPU capacity is provisioned on demand: when a GPU test needs a node, Karpenter provisions a real GPU instance from the pool's allowed families (a g4dn in practice, the cheapest) and releases it afterwards.

The e2e suite tests the default GPU pool on one instance family. Real hardware is only needed to prove the NVIDIA integration works (driver, device plugin, node fencing), and that part is the same on every family the pool allows; which family Karpenter picks is Karpenter's own optimization, so testing a second family would only re-test Karpenter and add capacity flakes. Custom pool configurations are covered by unit tests that compare the rendered configs to expected output, and workspaces requesting several GPUs are covered by the jupyter-k8s e2e tests on a fake-GPU node, where any GPU count can be tested for free.

### Implementation

1. `test_workspace_gpu.py` is the single mutating GPU module, five ordered cases in one `enable_default_gpu_pool` window: enable the pool (and check `jd pool list` plus the NVIDIA device plugin), the kubectl critical path (create from the GPU template, await Running, fenced-node checks, `nvidia-smi`, run the torch/CUDA notebook), the UI critical path (create from the "JupyterLab GPU" card, await via the UI, open, run the same notebook), scale back to zero, then disable the pool and verify the NodePool and `EC2NodeClass` are fully deleted rather than getting stuck mid-deletion (the regression in #349). If a test fails partway, a fixture turns `enable_default_gpu_pool` back to the value it had before the module ran, so the test deployment is left as it was found.
2. `JD_E2E_GPU_ENABLED` is the opt-in switch for the GPU tests, since they spend real GPU instance time. The CI e2e workflows now set it, so the GPU tests are included whenever CI runs e2e; for local runs you set it in `.env` (GPU tests need G/VT on-demand quota in the AWS account). The switch alone does not mean the deployment has a GPU pool (CI sets it against a deployment where the pool is off, and the GPU module enables the pool itself), so tests that need the pool also check the cluster and skip with a clear reason when it is not there.
3. GPU tests run as their own CI job ("Mutating tests: GPU", selected by the new `gpu` marker) after the general mutating job, so either can be re-run without repeating the other.
4. `test_autoscaling.py` adds a test that the routing services scale up under connection load and settle back down when the load ends: a helper deployment holds 600 connections open (3 pods x 200), enough to push the KEDA scaling metric over its thresholds, and the test asserts `traefik`, `authmiddleware`, and `web-app` grow above their pre-load replica counts and return to them after the connections close.
5. The shared test plugin (`pytest-jupyter-deploy`) gains the reusable pieces the new tests need: creating a workspace from a named template card in the web UI, a generic `poll` helper, a `resource_absent` check that only treats NotFound as absent, the connection ballast as a packaged manifest and context manager next to the existing CPU ballast, and the `gpu` marker.
6. Fixed argument quoting in the justfile (the repo's command scripts): the pytest marker and `-k` filter are now quoted through the `test-e2e` recipe, so multi-word expressions like `mutating and not gpu` survive the shell layers (previously a latent bug for any multi-word filter).

Out of scope of this PR: the GPU idle-shutdown chain (idle stop, then node reclaim) is not e2e-asserted, since the 60-minute idle default does not fit a run; it would fit a follow-up using a template at the 15-minute floor.


.github/workflows/e2e-eks-oidc-pr.yml
```diff
@@ -0,0 +1,68 @@
...
+# auto-creates the environment on first use with no rules and the gate passes
+# without review.
+#
+# Concurrency joins the e2e-eks-4 group shared with the manual and fresh
+# runs on OAuth app 4, so runs against the shared deployment serialize across
+# PRs and with manual dispatches.
+name: E2E PR (eks-oidc template)
+
+on:
+  pull_request:
```
JGuinegagne: does this mean we'll run the e2e tests on _any_ commit to the eks-oidc template?

If so, I would remove it, this will slow down the general CI too much.
Maintainer can still trigger the e2e tests manually against a specific PR (or ask their agent to do it).

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
```diff
@@ -33,19 +35,119 @@
...
+_CONN_BALLAST_CONNECTIONS_PER_POD = 200
+_TRAEFIK_IN_CLUSTER_HOST = f"traefik.{ROUTER_NAMESPACE}.svc.cluster.local"
+_PYTHON_IMAGE = "public.ecr.aws/docker/library/python:3.12-alpine"
+
+# Each holder pod keeps N TLS connections open to traefik's websecure
+# entrypoint, re-sending a keep-alive HEAD every 30s (under traefik's idle
+# timeout) and replacing dropped sockets, so the metric holds steady at the
+# target. Certificate verification is off: the in-cluster service DNS name is
+# not on the deployment's public certificate.
+_CONN_BALLAST_MANIFEST = string.Template(
```
JGuinegagne, line 57: anti-pattern: don't define yaml inline in python like this. Use [this plugin method](https://github.com/jupyter-infra/jupyter-deploy/blob/7283c07396c3d0126346217089e2527d1ce652f6/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_cluster_autoscaler.py#L50) instead (adapt as needed).

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_autoscaling.py
```diff
@@ -33,19 +35,119 @@
...
 
-def _poll(condition: "Callable[[], bool]", timeout_s: int, interval_s: int = 5, msg: str = "") -> None:
-    deadline = time.time() + timeout_s
-    while time.time() < deadline:
-        if condition():
-            return
-        time.sleep(interval_s)
-    raise TimeoutError(f"Condition not met within {timeout_s}s: {msg}")
+@contextmanager
+def _connection_ballast() -> Generator[None, None, None]:
```
JGuinegagne, line 133: remove please, use plugin method

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_golden_path.py
```diff
@@ -0,0 +1,113 @@
...
+from .conftest import WORKSPACE_NAMESPACE
+from .test_utils import NOTEBOOKS_DIR
+
+pytestmark = pytest.mark.usefixtures("kubernetes_cluster_login")
+
+PERSISTENCE_FLAG = "e2e-golden-path-flag.txt"
+
+
+@skip_if_testvars_not_set(["JD_E2E_USER", "JD_E2E_ORG", "JD_E2E_RBAC_TEAM"])
+def test_golden_path_workspace_journey(
```
JGuinegagne, line 28: isn't this essentially asserting that files written to home volume persist across restart?

If yes, then it's a duplicate, we already have coverage [here](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_home_volume.py).

We also have coverage for UI-driven app creation, stop, delete [here](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_web_app.py).

E2E tests are expensive to maintain, so let's not add duplicates please. Also, `golden_path` is not precise enough, so that's a bad practice.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_gpu_pool.py
```diff
@@ -0,0 +1,94 @@
+"""E2E test for the enable_default_gpu_pool on/off lifecycle on the EKS OIDC template.
+
+Unlike test_workspace_gpu.py, this does not need a GPU-enabled deployment: it
```
JGuinegagne, line 3: are we keeping [test_workspace_gpu](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py)?

If it needs an optional flag to run that the CI won't call, remove it altogether and update this comment please. However, we absolutely need coverage for gpu (both UI and `kubectl` path).

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_utils.py
```diff
@@ -0,0 +1,218 @@
...
+GPU_ROLE = "workspaces-gpu"
+GPU_ROLE_SELECTOR = f"jupyter-deploy/role={GPU_ROLE}"
+GPU_NODEPOOL = "workspace-gpu"
+# The karpenter-nodepools chart names each pool's EC2NodeClass after the pool.
+GPU_EC2NODECLASS = GPU_NODEPOOL
+
+
+def kubectl_stdout(*args: str) -> str:
+    result = subprocess.run(["kubectl", *args], capture_output=True, text=True, check=True)
+    return result.stdout.strip()
```
JGuinegagne, line 36: why do you need this and why isn't this contributed to the plugin [kubectl methods](https://github.com/jupyter-infra/jupyter-deploy/blob/7283c07396c3d0126346217089e2527d1ce652f6/libs/pytest-jupyter-deploy/pytest_jupyter_deploy/kubernetes/kubectl.py#L13) ?

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_utils.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def poll(condition: Callable[[], bool], timeout_s: int, interval_s: int = 5, msg: str = "") -> None:
```
JGuinegagne, line 39: move to plugin please

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_utils.py
```diff
@@ -0,0 +1,218 @@
...
+    deadline = time.time() + timeout_s
+    while time.time() < deadline:
+        if condition():
+            return
+        time.sleep(interval_s)
+    raise TimeoutError(f"Condition not met within {timeout_s}s: {msg}")
+
+
+def karpenter_resource_absent(resource: str, name: str) -> bool:
```
JGuinegagne, line 48: The methods in the file as a whole are an anti-pattern, I would remove this entire module.

`test_*` should run actual pytest test case, otherwise that's confusing.

Then at a minimum, we need as part of a mutating e2e tests:
- UI workflow for gpu (create via UI, await via UI, run GPU notebook)
- `kubectl` workflow for gpu (create w/ template, await, run GPU notebook)

Most likely though, you want to consolidate `test_gpu_pool` into a single mutating module (that flips the cluster flag), then run UI, `kubectl`, `pool` for gpu and node drain test cases.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -144,34 +52,6 @@ def test_gpu_workspace_kernel_sees_cuda(
...
-        delete_notebook(e2e_deployment, server_path, name=GPU_WORKSPACE, scope=WORKSPACE_NAMESPACE)
-    finally:
-        kubectl_delete_workspace(GPU_WORKSPACE)
-
-    _poll(
-        lambda: _gpu_node_count() == 0,
-        timeout_s=600,
-        msg="gpu NodePool did not scale to zero after workspace deletion",
-    )
+    verify_gpu_workspace_kernel_sees_cuda(e2e_deployment, dex_oauth_app)
```
JGuinegagne, line 57: anti-pattern, test is now pretty much empty and all the logic is in `test_utils`. Revisit please, but based on earlier comments this calls for a wider refactor.

scripts/env_setup_eks.py
```diff
@@ -54,6 +54,7 @@
     "team": "JD_E2E_TEAM",
     "rbac-team": "JD_E2E_RBAC_TEAM",
     "enable-default-gpu-pool": "JD_E2E_VAR_ENABLE_DEFAULT_GPU_POOL",
+    "gpu-tests": "JD_E2E_GPU_ENABLED",
```
JGuinegagne, line 57: remove this env var please

scripts/orphan_scan.py
```diff
@@ -0,0 +1,85 @@
...
+eventually consistent after deletions, so the scan runs in the caller's
+configured AWS region and retries before failing.
+
+Known blind spot: resources created at runtime by Kubernetes controllers (the
+NLB and its ENIs from the traefik Service, security groups from the load
+balancer controller) do not inherit terraform tags, so this scan cannot see
+them; cluster teardown deletes them via the Service/controller path instead.
+
+Usage: scripts/orphan_scan.py <deployment-id>
+"""
```
JGuinegagne, line 16: curious: what prompted you to add this?

While I agree that it's generally a good idea to scan for leftover resources, i/ I would recommend handling this in a separate PR (add an issue to track), and ii/ would move this to the plugin (it's generic).

JGuinegagne (changes requested):
High level: workspace `kubectl` and UI flows _must_ be exercised in e2e tests, as part of a mutating test. That _must_ include running a notebook that assert `cuda` is enabled. Consolidate into a single mutating module, and consider running as a separate job so that the GPU tests can be re-run separately (see base template for precedent)

andrii-i:
Thank you @JGuinegagne, all suggestions are adopted: consolidated the GPU coverage into one mutating module that runs both the kubectl and UI flows through the torch/CUDA notebook, moved the generic helpers into the pytest plugin, split the GPU tests into their own re-runnable CI job, removed the per-PR workflow and the golden-path test, and pulled the orphan scan out into #378.

andrii-i:
Running the e2e suites with this PR's test code: https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35121468416. Note that this doesn't run the new GPU job, as this PR needs to be merged for it to run. I verified the GPU module against a live deployment by running the pool lifecycle cases and walking the critical paths, both by hand and via an agent over MCP.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_golden_path.py
andrii-i: Good catch, removed. The one piece that wasn't duplicated (a notebook run on a UI-created workspace) is now part of the GPU UI case.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_gpu_pool.py
andrii-i: Merged them: one module (`test_workspace_gpu.py`) that enables the pool itself, and CI now sets `JD_E2E_GPU_ENABLED` so it runs in every e2e pass.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_utils.py
andrii-i: No good reason. Moved `poll` and a NotFound-only `resource_absent` into the plugin; the kubectl calls now go through the plugin's `run_kubectl`.

scripts/env_setup_eks.py
andrii-i: Removed. CI opts in via an explicit step in the job workflow instead; locally it remains a manual `.env` opt-in.

scripts/orphan_scan.py
andrii-i: The stuck teardown in #349 prompted it. Pulled it out of this PR entirely; opened #378 to re-land it as a plugin module like you suggest.

andrii-i:
CI run from a branch in the upstream (original) repo: https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35161657837

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_components.py
```diff
@@ -254,6 +256,7 @@ def test_nvidia_device_plugin_daemonset_when_gpu_enabled(e2e_deployment: EndToEn
     non-GPU deployments).
     """
     e2e_deployment.ensure_deployed()
+    require_gpu_pool()
```
JGuinegagne: non-blocking: it's odd to define a plain util in `conftest.py`; that module is usually reserved for pytest fixtures.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -1,90 +1,150 @@
...
+# Display name of the default GPU WorkspaceTemplate the flag synthesizes
+# (engine/platform_karpenter.tf), rendered as a card on the create page.
+GPU_TEMPLATE_DISPLAY_NAME = "JupyterLab GPU"
 GPU_ROLE = "workspaces-gpu"
 GPU_ROLE_SELECTOR = f"jupyter-deploy/role={GPU_ROLE}"
-GPU_NODEPOOL = "workspace-gpu"
+# The karpenter-nodepools chart names each pool's EC2NodeClass after the pool.
+GPU_EC2NODECLASS = GPU_NODEPOOL
+
+ORDER_GPU = 10
```
JGuinegagne: for consistency w/ base template e2e tests, consider moving to `tests/e2e/constants.py`

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -1,90 +1,150 @@
...
+    _apply_gpu_pool_flag(e2e_deployment, True)
+
+    assert gpu_pool_deployed(), "workspace-gpu NodePool missing after enabling enable_default_gpu_pool"
+
+    result = e2e_deployment.cli.run_command(["jupyter-deploy", "pool", "list"])
+    assert GPU_NODEPOOL in result.stdout, f"Expected pool '{GPU_NODEPOOL}' in pool list output:\n{result.stdout}"
+
+    daemonset = run_kubectl(
+        "get", "daemonset", "nvidia-device-plugin", "-n", "kube-system", "-o", "jsonpath={.metadata.name}"
+    )
```
JGuinegagne, line 115: could you use the `jd component status` & target the `plugin` daemonset here?

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -96,82 +156,120 @@ def test_gpu_workspace_provisioning_nvidia_smi_and_scale_to_zero(e2e_deployment:
...
+
+    poll(
         lambda: _gpu_node_count() == 0,
         timeout_s=600,
         msg="gpu NodePool did not scale to zero after workspace deletion",
     )
+
+
+@pytest.mark.order(ORDER_GPU + 4)
+@skip_if_testvars_not_set(["JD_E2E_GPU_ENABLED"])
```
JGuinegagne, line 250: sanity-check: shouldn't we remove this?

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -96,82 +156,120 @@ def test_gpu_workspace_provisioning_nvidia_smi_and_scale_to_zero(e2e_deployment:
...
+        delete_notebook(e2e_deployment, server_path, name=name, scope=WORKSPACE_NAMESPACE)
     finally:
-        kubectl_delete_workspace(GPU_WORKSPACE)
+        # UI deletion is covered in test_web_app; deleting via kubectl avoids the duplicate.
+        kubectl_delete_workspace(name)
+
 
-    _poll(
+@pytest.mark.order(ORDER_GPU + 3)
+@skip_if_testvars_not_set(["JD_E2E_GPU_ENABLED"])
```
JGuinegagne, line 236: sanity-check: shouldn't we remove this?

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
```diff
@@ -1,90 +1,150 @@
...
 def _gpu_node_count() -> int:
     return len(get_node_names(GPU_ROLE_SELECTOR))
 
 
+def _apply_gpu_pool_flag(e2e_deployment: EndToEndDeployment, enabled: bool) -> None:
+    """Record the flag in variables.yaml, then apply with jd config + jd up.
+
+    jd config exposes the variable only as a set-true flag (typer generates no
+    --no- form for explicitly named bool options), so both directions go
+    through the variables.yaml override.
```
JGuinegagne, line 69: fine for now, but we need to find a way to handle boolean values properly in `jd config`. Could you create a GH issue to track this?

JGuinegagne (dismissed):
Great improvements, thanks Andrii!

LGTM, w/ non-blocking comments & suggestion.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_components.py
andrii-i: Converted it to a gpu_pool_required fixture; conftest now holds only fixtures.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Done, moved the GPU constants to tests/e2e/constants.py.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Not yet: the manifest doesn't declare a nvidia-device-plugin component, so component status has nothing to target, and declaring one today would break test_component_status_happy_case on GPU-off deployments (it polls every declared DaemonSet to Ready, and components can't be conditional). Opened #380 for flag-gated components; keeping the kubectl check until that lands.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Kept it deliberately: without this gate, a mutate=true run without the GPU opt-in against a deployment whose owner enabled the pool would run this case and fail on their live GPU workspaces (nodes never reach zero).  So this env var protects local runs from failures that have nothing to do with the code under test.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Same reasoning, with a sharper edge: ungated, this case would disable a pool the owner turned on deliberately, draining any GPU node mid-use, then the guard would flip it back, two applies of churn on a cluster the run had no GPU intent for.

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Opened #379.

andrii-i:
CI run from the upstream branch with the latest changes https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35259576677

libs/jupyter-deploy-tf-aws-eks-oidc/tests/e2e/test_workspace_gpu.py
andrii-i: Kept for the same reasons as https://github.com/jupyter-infra/jupyter-deploy/pull/366#discussion_r4040213434 but here it would be worse: case turns the pool off so it would kill the owner's running GPU workspaces mid-use.

JGuinegagne (approved):

PR #369, opened 2026-09-03
chore: jupyterlab ci

JGuinegagne:
This PR setups the core CI for the `tf-aws-ec2-jupyterlab` template, mirroring base/eks template CIs: a/ a `release-jupyterlab` workflow (build → lint → test → Test PyPI → live e2e gate → PyPI); b/ a 3-job e2e workflow (deploy+verify → test → destroy) that fresh-deploys in-container and tears down every run without using persistent slot, since jupyterlab needs only AWS creds; c/ cuts the template to `0.1.0rc1`. Jupyterlab has no OAuth app / domain, so e2e runs scope by deployment id and get a dedicated ECR trio (repos 7-9, one per concurrent trigger).

Also in this PR:
- fix a rare IAM propagation race: a `time_sleep` gate holds the EC2 instance until its instance profile is visible to `RunInstances` (was flaking with "Invalid IAM Instance Profile name")
- broaden `tests/unit/test_version.py` service-file regex to accept pre-release versions (`rc`/`a`/`b`)
- add the `ec2-jupyterlab` PyPI pill to the root README

### Developer experience
- `release-jupyterlab` (`workflow_dispatch`) publishes to Test PyPI then PyPI, gated by a live fresh-deploy e2e
- `e2e-jupyterlab-fresh` (`workflow_dispatch` + `workflow_call`) fresh-deploys, tests, and destroys; per-slot ECR (7=PR/dispatch, 8=release, 9=canary) so triggers run in parallel
- `just update-version jupyterlab-template <patch|minor|major|version>` bumps the template (all 8 version files)
- `just test-e2e-jupyterlab` / `ci-e2e-jupyterlab-deploy` / `ci-restore-jupyterlab` / `find-takedown-jupyterlab` for local + CI deploy, restore, and teardown

### Implementation
1. `.github/workflows/release-jupyterlab.yml`: build → lint → test → Test PyPI → `e2e-jupyterlab-release` gate → PyPI → GH release; `prerelease` auto-flagged for rc/a/b
2. `.github/workflows/e2e-jupyterlab-fresh.yml`: 3 jobs — deploy+verify (`test_deployment`), test (non-deployment), destroy; per-slot concurrency `e2e-jupyterlab-<slot>` so release/canary/PR run in parallel
3. `.github/workflows/e2e-jupyterlab-job.yml`: reusable test job — restores the deploy from the store by deployment id, runs the suite (`full_deployment` auto-skips on an existing project)
4. `.github/workflows/e2e-jupyterlab-release.yml`: release-gate wrapper → fresh (pypi install, slot 8)
5. `justfile` + `scripts/ci_restore_jupyterlab.py` / `find_takedown_jupyterlab.py`: the in-container deploy emits `deployment_id`; restore + takedown scope to it (base/eks match by subdomain — jupyterlab has none); takedown regenerates config before `jd down`, then deletes the store entry
6. `.github/e2e-jupyterlab/pyproject.release.toml`: template from Test PyPI + CLI `[aws,proxy]` from the workspace (temporary — the `[proxy]` extra can't resolve from PyPI until a final CLI/proxy ships)
7. `.github/workflows/e2e-build-image.yml` + `.github/e2e-shared/Dockerfile`: `TEMPLATE=jupyterlab` selects the release pyproject
8. `libs/jupyter-infra-tf-aws-iam-ci/` (`ecr.tf`/`outputs.tf`): add ECR repos 7-9, generically (no OAuth tags) so a third template gets one repo per concurrent trigger
9. `scripts/upgrade_jupyterlab_template_version.py` + `scripts/update_version.py`: `jupyterlab-template` target bumps all 8 version files
10. `template/engine/modules/ec2_iam_role`: `time_sleep` trigger carries the instance-profile name so the instance waits for IAM propagation

### Testing
- ci: full chain green end-to-end — deploy → test → destroy, no orphaned resources ([run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/33796481012))
- ci: the `release-jupyterlab` e2e gate runs the same fresh-deploy path installing the template from Test PyPI
- manual: exercised `find-takedown-jupyterlab` locally (restore by id → regenerate config → `jd down` → delete store entry); confirmed the `time_sleep` fix clears the `RunInstances` race


bhanderig (approved):

.github/e2e-jupyterlab/pyproject.release.toml
```diff
@@ -0,0 +1,36 @@
...
+[[tool.uv.index]]
+name = "extra"
+url = "{{EXTRA_INDEX_URL}}"
+explicit = true
+
+[tool.uv.sources]
+jupyter-deploy-tf-aws-ec2-jupyterlab = { index = "extra" }
+pytest-jupyter-deploy = { path = "libs/pytest-jupyter-deploy", editable = true }
+# TEMPORARY (see header): CLI + proxy from the workspace until a final CLI/proxy is on PyPI.
+jupyter-deploy = { path = "libs/jupyter-deploy", editable = true }
```
andrii-i, line 35: With the CLI installed from a path, the pypi-mode `uv sync` fails on the CLI's own [`jupyter-deploy-client-proxy = { workspace = true }`](https://github.com/jupyter-infra/jupyter-deploy/blob/560c2a228f12f8a415c7c01e810cb0e057f57f8f/libs/jupyter-deploy/pyproject.toml#L60-L63) source, which uv only accepts inside a workspace (building the E2E image with the release gate's build args reproduces it), so the first `release-jupyterlab` dispatch would fail at the image build. We should make this file a `[tool.uv.workspace]` with the three libs as members and `workspace = true` sources; that shape builds. Stripping the CLI's sources instead does not resolve, since the proxy is `0.1.0rc1` and the `[proxy]` extra requires `>=0.1.0`, which uv only waives for workspace members.

andrii-i (approved):
LGTM, one thing worth looking into below as it would fail the `release-jupyterlab` run.

PR #373, opened 2026-09-14
test: jupyterlab template e2e

JGuinegagne:
This PR takes the `tf-aws-ec2-jupyterlab` e2e suite from a thin smoke test to coverage parity with the base template covering a/ the parts the base template cannot have (the client proxy, the STS-identity auth boundary, both `jd open` modes), b/ the first mutating pass for this template: two in-place `jd config` + `jd up` cycles on one deployment, and c/ the plugin utilities they need. Writing the tests surfaced five product bugs, fixed here.

Also in this PR:
- the auth sidecar now enforces the token's *signed* lifetime (`X-Amz-Date + X-Amz-Expires`), which STS itself does not — a captured token's replay window drops from STS's ~15min SigV4 skew to the 60s the proxy minted it for
- `jd teams`/`jd users` no longer return before the new allowlist is live: the sidecar healthcheck drops 30s → 5s so `docker compose up --wait` is a real gate, and a sidecar that never comes back healthy now fails the command instead of being logged and ignored

### User experience
- `jd proxy stop` no longer reports "No running proxy found" (non-zero) for a proxy it just stopped correctly
- `jd proxy status` no longer reports a SIGKILLed proxy as running, and `jd proxy start` no longer refuses to replace it
- `jd open` / `jd proxy connect-info` against a stopped host now say what to do (`jd host start`) instead of printing a traceback
- `jd server exec -s <unsupported-service>` now names the problem instead of surfacing a botocore traceback
- `jd teams add|remove` returning 0 now means the access decision is already in effect

### Developer experience
- run the whole suite against a fresh deployment: `just test-e2e-jupyterlab`
- run against an existing deployment: `just test-e2e-jupyterlab <project-dir>`
- include the two-apply mutating pass: `just test-e2e-jupyterlab <project-dir> "" mutate=true`

### Implementation
1. `<plugin>/local_proxy/requests.py`: pinned/unverified/system-trust TLS request helpers, cert-pin inspection, and `static_bundle_proxy` (a proxy driven from a fixed bundle)
2. `<plugin>/local_proxy/jupyterlab.py`: Jupyter's HTTP/WS API through the proxy — XSRF-bearing session, kernel start/list/delete, `kernel_websocket` negotiating the v1 kernel subprotocol
3. `<plugin>/local_proxy/application.py`: `verify_app_status(expected)`, the negative counterpart of `verify_jupyterlab_accessible()` — a browser cannot otherwise tell 403 from "did not load"
4. `<plugin>/auth_sidecars/aws_auth_sidecar.py`: client half of the sidecar's token contract — mint, decode, and deliberately malform `k8s-aws-v1.` tokens; docstring-only `__init__` keeps `boto3` out of bare installs
5. `<plugin>/plugin.py` + `cli.py` + `suite_config.py` + `pyproject.toml`: proxy-mode fixtures, `jd proxy` subcommand wrappers, `strip_ansi` for pty transcripts, explicit `websocket-client` dep
6. `<lab-template>/tests/e2e`: `conftest`/`constants` (ordering ordinals + the mutating pass's mount and flag constants) plus notebook and file assets
7. `<lab-template>/tests/e2e`: 13 modules — config/pre-deploy, deployment/host/home-volume, proxy/open/application, auth/teams/users, server/show — and the two-apply mutating pass (`uv` → GPU+`pixi`+external volumes → CPU+`uv`) asserting what survives each instance replacement: additional data volume, cert pin, external mounts
8. `<lab-template>/services/auth-sidecar/main.go`: bound `X-Amz-Expires` to ≤900s and reject a token past `X-Amz-Date + X-Amz-Expires`, ±60s clock skew
9. `<lab-template>/services/commands/update-allowlist.sh` + `docker-compose.yml.tftpl`: healthcheck `interval: 5s`; exit non-zero on an unhealthy sidecar, after the terraform write-back so a failed wait cannot split-brain the allowlist
10. `<cli>` fixes: zombie PIDs no longer count as alive (`cmd_utils`), "running" requires both a live status and a verified process identity (`proxy/proxy_manager`), and typed hinted errors replace bare `ValueError`/`ClientError` (`api/aws/ec2/ec2_instance`, `api/aws/ssm/ssm_command`)

### Testing
- e2e: full non-mutating suite against a live deployment — 92 passed / 10 skipped
- e2e: two-apply mutating pass (GPU swap and back) against a live deployment
- e2e: `test_open`, `test_teams`, `test_users`, `test_start_server` re-run individually after the assertion rework, all green
- manual: verified the sidecar healthcheck landed at 5s on the instance after `jd config` + `jd up`; confirmed `jd teams remove` returns only once the revoke is live (the tests assert the edges once, with no polling)
- unit: `<cli>` fixes each land with their unit tests (zombie reaping, proxy status/identity, typed AWS errors)


JGuinegagne:
Full e2e test [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/34903886497) against sibling branch

JGuinegagne:
E2E [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/34909653378) that exclude `tornado` induced [regression](https://github.com/tornadoweb/tornado/issues/3724)

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/configurations/env.example
```diff
@@ -15,3 +15,23 @@
...
+# Note log_files_retention_days rather than volume_size_gb: AWS enforces a 6-hour
+# cooldown between EBS volume modifications, so a volume resize cannot be re-run.
+
+JD_E2E_CPU_INSTANCE=t3.medium
+# g5 rather than g6: the templates pin the instance to `subnet_ids[0]` of the default VPC, so the
+# GPU swap can only ever be attempted in that one AZ. On 2026-09-15 g6.xlarge was refused there in
+# both accounts we run in — us-west-2c (dev) and us-west-2d (CI), 0/4 attempts, ~52min of provider
+# retries per attempt — while g5.xlarge got capacity in about a minute. Revisit once the AZ is
+# selectable rather than positional (jupyter-deploy#374).
+JD_E2E_GPU_INSTANCE=g5.xlarge
```
andrii-i, line 40: With this default the mutating job cannot pass in the CI account: the template pins the instance to the first default-VPC subnet, `us-west-2d` there, and EC2 does not offer `g5.xlarge` in that zone ([failed run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35009728619/job/104538116393)). PR checks do not run this workflow, so this surfaces only when the [release workflow](https://github.com/jupyter-infra/jupyter-deploy/blob/983f2b6a6058cfe1934b5e02e74a4d8e08b91fd1/.github/workflows/release-jupyterlab.yml#L149) runs it as the gate for `prod-pypi-publish`.

Until #374 makes the zone selectable instead of positional, we should pass the mutating job a type the CI account offers in `us-west-2d` (`aws ec2 describe-instance-type-offerings --location-type availability-zone --filters Name=location,Values=us-west-2d` with the CI credentials; `just env-setup-jupyterlab gpu-instance=<type>` in the workflow's seed step keeps this default for dev) or keep the job out of the release gate. Fine as a follow-up if no release is planned before that.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main_test.go
```diff
@@ -0,0 +1,18 @@
+package main
+
+// No unit tests yet — deliberately.
```
andrii-i, line 3: This has no test functions, while the PR description's Testing section says `main_test.go` covers the sidecar's expiry and skew bounds. With the file present `go test` reports `ok ... [no tests to run]` for the package instead of `[no test files]`. Should we either add tests now or remove this file, update the `test_auth.py` docstring that points at it, and drop the unit-test claims from the PR description?

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/commands/update-allowlist.sh
```diff
@@ -131,13 +135,30 @@ if [ "$CURRENT" != "$FINAL" ]; then
...
-    log_message "auth-sidecar recreate returned non-zero (change is applied to .env regardless): $OUTPUT"
+    # Roll the host edit back, because this command is about to fail and a failed command gets NO
+    # terraform write-back: aws_ssm_runner raises on SSM Status=Failed before returning results, and
+    # the handlers only call update_variables() on success. Leaving the edit in place would put the
+    # host ahead of terraform, and the direction that costs is `remove` — the host would have
+    # revoked the entry while the terraform variable still lists it, so the next `jd up` re-seeds
+    # /etc/AUTH_ALLOWLIST and silently restores the access the operator just revoked.
+    # Rolling back keeps both sides at the pre-command state, so a non-zero exit means exactly
+    # "nothing changed, retry once the sidecar is healthy".
+    log_message "ERROR: auth-sidecar did not become healthy within 60s: $OUTPUT"
```
andrii-i, line 152: `log_message` writes only to the host log, so on this path the CLI receives SSM `Status=Failed` with an empty stderr and [prints](https://github.com/jupyter-infra/jupyter-deploy/blob/983f2b6a6058cfe1934b5e02e74a4d8e08b91fd1/libs/jupyter-deploy/jupyter_deploy/cli/error_decorator.py#L209-L218) `Command '<doc>' failed` and the rolled-back list. Should we clearly say that the sidecar did not come back healthy and the edit was rolled back? Writing it to stderr would show it in the CLI output and still keep it in the host log.

andrii-i (dismissed):
Thanks @JGuinegagne, coverage expansion and fixes look good. Note the comment on `env.example`: with the default `JD_E2E_GPU_INSTANCE` the mutating job fails because the CI account's pinned zone (`us-west-2d`) does not offer `g5.xlarge`, and the job sits in the release gate, so the next template release would fail there. Fine as a follow-up if no release is planned before that. Other comments are nonblocking.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main_test.go
JGuinegagne: Thanks Andrii. I updated the PR body to stop claiming there are go test.

`test_auth.py` mentions that we don't have tests yet, so I would say we're good.
Ack re `go test`, but i think we can live w/ this till we migrate the auth sidecar to its own repo; JD is a python repo, devs / CIs should not run `go` commands in it.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/configurations/env.example
JGuinegagne: Already addressed as I was hitting repeated ICE issues; now pinning AZ to `us-west-2a` in the CI via the flag.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
```diff
@@ -40,6 +40,22 @@ variable "instance_type" {
...
 
+variable "availability_zone" {
+  description = <<-EOT
+    The availability zone to place the EC2 instance and its EBS volumes in.
+
+    Use "any" to accept the first subnet of the default VPC, whichever zone that happens to be.
+    Set it when the instance type you want has no capacity in that zone: an instance type is not
+    offered in every zone, and even where offered it can be refused for lack of capacity.
+
+    Changing this on an existing deployment is NOT supported: EBS volumes cannot cross zones, so
```
andrii-i, line 51: Should we add an enforcement mechanism for this, seeing how changing the zone on an existing deployment replaces the home data volume and loses its data? `jd config` could proactively fail with a clear message when the deployment already has a data volume in another zone at plan time. Alternatively pinning the volume to its zone with `ignore_changes` is a smaller fix but it would fail only at attach time after the instance has moved, leaving the user with a fresh instance in a new zone and an unattached data volume, which is worse user experience.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_show.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+import re
+
+import pytest
+from jupyter_deploy.enum import ValueSource
+from pytest_jupyter_deploy.deployment import EndToEndDeployment
+from pytest_jupyter_deploy.terraform.utils import get_variables_dot_tf_path
+
+_SENSITIVE_VARIABLE_RE = re.compile(r'variable\s+"(?P<name>[^"]+)"\s*\{(?P<body>[^}]*)\}', re.DOTALL)
```
andrii-i: `[^}]*` stops at the first `}`, so when a variable has a `validation` block a `sensitive = true` written after it is never seen (and most variables in this template have one). Reproduced by appending to `variables.tf`: the regex reports nothing and the test passes.

The `parse_variables_dot_tf_content` function in `jupyter_deploy.engine.terraform.tf_varfiles` already returns each variable with its `sensitive` flag and reports the injected variable. Should we use that here instead of the regex?

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main.go
```diff
@@ -51,9 +53,14 @@ const (
 	tokenPrefix       = "k8s-aws-v1."
 	bindingHeader     = "x-k8s-aws-id"
 	maxTokenExpirySec = 900
-	cacheTTL          = 60 * time.Second
-	stsCallTimeout    = 5 * time.Second
-	stsRetryBackoff   = 200 * time.Millisecond
+	// Tolerance for clock skew between the signing client (the user's laptop) and this host, applied
+	// in both directions when enforcing the token's signed lifetime.
+	maxClockSkew = 60 * time.Second
```
andrii-i, line 58: The slack is measured against the server clock, and the proxy mints 60s tokens refreshed at 45s, so a client clock more than about a minute fast or slow has every token rejected with a bare 401, and the reason is only in the sidecar log, with the user seeing an unauthorized page with nothing that points at the clock. Tokens are stamped by the local clock and checked against the server's, so any local clock error over a minute becomes a lockout, and local VMs are prone to exactly that: [WSL2](https://github.com/microsoft/WSL/issues/5324) and [Docker Desktop](https://github.com/docker/for-mac/issues/17) both fall behind by the length of a host sleep until they re-sync, which is also how the E2E container runs `jd` on our Macs. The closest precedents allow more: EKS, which validates this same token format, [accepts 15 minutes](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/5ae0e8e3b6dea279c25278c8f4e701cd3f7bcbb7/pkg/token/token.go#L99), and Kerberos, the classic time-based login, [defaults to 5](https://web.mit.edu/kerberos/krb5-latest/doc/admin/conf_files/krb5_conf.html). Should we allow 5 minutes, which still caps a stolen token at about 6 minutes against the 15 STS itself gives, and have the 401 say the token is outside its time window so the user can check their clock? Fine as a follow-up.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_host.py
```diff
@@ -0,0 +1,296 @@
...
+def test_host_stopped_fails_open_gracefully(host_cycle: HostCycleObservations) -> None:
+    """`jd open` against a stopped host fails with an actionable error and never auto-starts it.
+
+    Starting an instance is a billable side effect, so `jd open` must not do it implicitly —
+    the user asked to open the app, not to provision. It reports the problem and stops.
+    """
+    assert host_cycle.stopped_open_failed, "`jd open` did not fail gracefully against a stopped host"
+    assert host_cycle.running_host_status == "running", (
+        "Sanity: the host must have been started by `jd host start`, not by `jd open`"
+    )
```
andrii-i: This reads the host status after `jd host start`, so it passes whether or not `jd open` started the host. Should we read the status right after the failed `jd open` instead?

justfile
```diff
@@ -875,31 +883,30 @@ ci-e2e-jupyterlab-deploy project_dir="sandbox-e2e":
     } > "$OVERRIDE_FILE"
     trap 'rm -f "$OVERRIDE_FILE"' EXIT
```
andrii-i: Remove this block? `test-e2e` writes the same override and restarts the container itself.

libs/pytest-jupyter-deploy/pytest_jupyter_deploy/image/Dockerfile
```diff
@@ -61,6 +61,13 @@ RUN set -e; \
 WORKDIR /workspace
 RUN chown ${USER_UID}:${USER_GID} /workspace
 
+# Shared terraform provider plugin cache (TF_PLUGIN_CACHE_DIR, set in docker-compose.yml).
+# Without it every `terraform init` — one per `jd config`, and the config tests run many in
+# throwaway /tmp projects — re-downloads the ~140 MB AWS provider into its own .terraform/.
+# Terraform requires the directory to exist; it never creates it.
+RUN mkdir -p /home/testuser/.terraform.d/plugin-cache && \
+    chown -R ${USER_UID}:${USER_GID} /home/testuser/.terraform.d
```
andrii-i: Remove this? The cache is the host `${HOME}` path that compose bind-mounts, so this directory is never used.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/test_teams.py
```diff
@@ -0,0 +1,275 @@
...
+from pytest_jupyter_deploy.deployment import EndToEndDeployment
+from pytest_jupyter_deploy.local_proxy import LocalProxyApplication
+
+# Every test here edits the live allowlist, so the snapshot/restore net is mandatory: a failure
+# between a revoke and its restore would 403 every later test in the file.
+pytestmark = pytest.mark.usefixtures("restore_allowlist")
+
+# `jd teams` recreates only the auth-sidecar container, during which Traefik's ForwardAuth can
+# briefly see a connection refused and answer 5xx. Poll to the expected status rather than sampling
+# once — the claim under test is "the new decision takes effect", not "within one round-trip".
```
andrii-i: Remove this?

andrii-i (dismissed):
Thanks for addressing the comments @JGuinegagne. Please find some more comments below, sorry for the second round. The one to look at first is on `variables.tf`, changing the zone on an existing deployment replaces the data volume and we have no enforcement mechanism to protect users from data loss. The rest are nonblocking.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/configurations/env.example
andrii-i: Confirmed, the 11b9706 run passed the mutating pass (8 passed). Thanks.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main_test.go
andrii-i: Makes sense, thanks for fixing the description. Worth an issue so the unit tests land with the extraction?

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/commands/update-allowlist.sh
JGuinegagne: thanks, that would be a good UX improvement, but we should address it across all host commands.
Tracking issue: #376

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
JGuinegagne: I think we should use EBS snapshots to guard against data loss on EBS volumes (home volume and additional EBS mounts) when switching AZs. That's a larger effort though, so I'll take it in a subsequent PR.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/services/auth-sidecar/main.go
JGuinegagne: makes sense, I haven't observed this in the multiple e2e test runs, but this indeed sounds like the right tradeoff between security and availability. I'll take it as a follow up.

JGuinegagne: actually, on second thought, I don't think we should loosen security (token validity) on a theoretical clock drift issues; thanks for the precedents though, very helpful. TLDR: let me sleep on this. Updating the error message now to provide users a debug path.

JGuinegagne:
Successful E2E run: https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35089585263

Retriggering w/ the latest changes.

andrii-i (approved):

JGuinegagne:
Full e2e test run: https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35105088902

PR #381, opened 2026-09-18
fix: az volume backup

### LINKED ISSUE #374: Instance swaps fail deterministically on AZ-pinned InsufficientInstanceCapacity
Both EC2 templates pin the instance to the first subnet of the default VPC, so a scarce instance type fails deterministically in one AZ with no way out.

- `jupyter-deploy-tf-aws-ec2-base/.../engine/main.tf:85` and `jupyter-deploy-tf-aws-ec2-jupyterlab/.../engine/main.tf:88` both set `subnet_id = module.network.subnet_ids[0]`, where `subnet_ids` is `data.aws_subnets.default_vpc_subnets.ids`. Neither template exposes an AZ or subnet variable, and neither has capacity handling.
- The subnet order is whatever the EC2 API returns — arbitrary but stable per account. In one dev account `[0]` is `us-west-2c`; another account lands elsewhere, so the blast radius differs per account and is invisible until an apply fails.

**Impact**
- `g6.xlarge` in `us-west-2c` was refused repeatedly on 2026-09-14 (failed after 52 min) and 2026-09-15 (~16 min of retries), while `g5.xlarge` got capacity instantly in the same AZ. `g6.xlarge` is *offered* in all four `us-west-2` AZs — this is pure capacity.
- The AWS provider treats `InsufficientInstanceCapacity` as retryable, so `terraform apply` sits in `Still creating...` with no error surfaced. It reads as a hang, not a failure.
- Terraform destroys the old instance before creating the new one, so a refused swap leaves the deployment with **no instance**.
- The home volume lives in the instance's AZ and EBS cannot cross AZs, so an existing deployment cannot be moved to chase capacity — only a fresh deploy can.
- Hits `test_gpu` (base) and `test_mutating_gpu_pixi` (jupyterlab) on every GPU run, and any user who upgrades their instance type.

**Proposal**
- Expose an `availability_zone` (or `subnet_id`) variable in both templates, defaulting to today's behaviour so existing deployments are unaffected.
- Optionally, when the variable is unset, select the subnet whose AZ offers the requested instance type via `aws_ec2_instance_type_offerings` instead of taking `[0]` blindly.
- Document that changing the AZ of an existing deployment requires a fresh deploy (EBS is AZ-bound).

Workaround today: pick a type with capacity in the pinned AZ (`JD_E2E_GPU_INSTANCE=g5.xlarge` for the E2E suites).

JGuinegagne:
Changing `availability_zone` on a deployed `tf-aws-ec2-jupyterlab` project replaced the home data volume and silently lost its contents: EBS volumes cannot cross zones, and nothing warned, blocked, or backed up. This PR adds a/ `jd volume` (`list`/`show`/`status`/`backup`) over a new template-declared `volumes:` manifest section, and b/ `jd config --restore-volumes`, which resolves each volume's latest backup into a terraform variable so the next `jd up` recreates it from that backup instead of empty. Taking the backup is a separate, explicit command because planning must not create billable resources.

Also in this PR:
- e2e CI: the GPU apply moves to its own blocking job so "re-run failed jobs" retries only it, teardown runs only when every test job passed, and a scheduled reaper (`e2e-reap-stale.yml`) destroys any jupyterlab e2e deployment older than 12h
- `jd config` could not write a map whose keys are not HCL identifiers: `format_terraform_value` emitted bare keys, so `home/external-ebs1` failed the plan with `Variables not allowed`
- dropped `test_allowlist_write_back_leaves_no_terraform_diff`: asserting an empty plan fails on any unrelated drift, e.g. the AL2023 AMI alias moving under a live deployment ([run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35287268310), twice); `test_teams_list_matches_the_terraform_variable` covers this.

Closes: #374 — it proposed exposing an AZ variable and documenting a fresh deploy as the only way to change it; this makes the change non-destructive instead. `tf-aws-ec2-base` carries the same exposure and is not covered here (separate port).

### User experience
- inspect the storage backing an app: `jd volume list` (Name / Class / Description), `jd volume show --name <volume-name>`, `jd volume status [--name <volume-name>]`
- back up a volume: `jd volume backup --name <volume-name>` / `--all`, refused unless the host is stopped, since a snapshot of a live disk can restore corrupt
- survive a zone change with the data intact: `jd volume backup --all` → `jd config --restore-volumes --availability-zone <zone>` → `jd up`
- volumes are named by mount path as seen in the app (`home`, `home/<mount-point>`), not by provider id or terraform resource name
- EFS mounts are listed but report `regional` / `elastic` / `not needed`: a zone change relocates their mount target rather than destroying them
- refusals name the reason and the remedy, and are distinct: a volume the deployment only references, a class with no backup mechanism, a backup older than the last session, a name that is not mounted (which lists the ones that are)
- `--restore-volumes` refuses unless every volume has a backup taken after the host last stopped, so a restore cannot silently discard a session's work

### Implementation
1. `<lab-template>/engine/variables.tf` + `presets` + `variables.yaml`: `ebs_snapshot_ids` map, threaded into `modules/volumes` as `snapshot_id` on the home volume and each `additional_ebs_mounts` entry
2. `<lab-template>/engine/modules/volumes`: volume identity composed as `home` / `home/${mount_point}` next to the lookup, so the tag, the key and the declared name cannot drift
3. `<lab-template>/engine/modules/volumes/main.tf`: `when = destroy` provisioner deletes every snapshot carrying the deployment id
4. `<lab-template>/engine/outputs.tf`: `availability_zone` plus jsonencoded `additional_ebs_volumes` / `additional_efs_volumes` inventories, one entry per configured mount
5. `<lab-template>/manifest.yaml`: `volumes.static` / `volumes.dynamic` declarations and the `volume.*` commands; `volume.list` branches on the storage class inside one command, so a new kind adds a branch rather than a sibling command
6. `<cli>/manifest.py`: `volumes:` schema, well-known `volume.list` / `volume.validate-backups-ready` command names whose presence is the opt-in, and a validator rejecting a duplicated static name (it keys the backups map)
7. `<cli>/handlers/resource/volume_handler.py` + `cli/volume_app.py`: inventory join, per-kind live-state lookup, backup ordering (create → wait `completed` → delete the superseded one), and the typed refusals
8. `<cli>/api/aws/ec2/{ebs_snapshot,ebs_volume}.py` + `api/aws/efs/` + `provider/aws/aws_efs_runner.py`: the only boto3; EFS is queried by id since `DescribeFileSystems` has no tag filter
9. `<cli>/api/aws/ec2/ec2_instance.py`: quiescence gates — the host's last stop parsed from `StateTransitionReason`, and a refusal when any backup predates it or the stop time is unknown
10. `<cli>/handlers/project/config_handler.py`: `restore_volumes()` validates before writing, then writes the resolved map through both `update_variable_records` and the config sync, as `updates:`-declaring commands do
11. `docs/source/reference/resource`: `volume` registered in the resource group, which a new command group needs to get a page at all; `AGENT.md` records the regeneration step
12. `<plugin>/volumes.py` + `plugin.py`: `get_volume_names` / `get_volume_details` helpers, a module-scoped `stopped_host` fixture that yields *while* stopped, and a collection guard rejecting `cli` + `mutating` on one test
13. `<lab-template>/tests/e2e/test_volume.py`: non-mutating, home volume only, every command reached without `--name` so the default resolution is itself under test
14. `<lab-template>/tests/e2e/test_volume_swaps.py`: mutating, ordered last; `volumes_provisioned` provisions the EBS + EFS mounts and pins the base zone, re-resolving the backups map only when a zone change is pending
15. `.github/workflows/e2e-jupyterlab-fresh.yml` + `e2e-reap-stale.yml` + `scripts/reap_stale_jupyterlab.py`: GPU apply isolated, teardown gated on every test job passing, and the reaper scoped by caller credentials, `Template` tag and age

### Testing
- e2e: full suite against a fresh deployment, 123 passed / 0 skipped, including the zone swap that writes marker content before the change and reads it after
- e2e: `test_a_backup_older_than_the_last_session_is_refused`, which manufactures the no-symptom case — back up, run the app, stop it again — and asserts the refusal names the offending instant
- e2e: both mutating applies (`uv`/CPU → `pixi`/GPU → `uv`/CPU) re-run after the volume work

### Screenshots
**jd volume --help**
volume-help

**jd volume list**
volume-list

**jd volume show (home volume)**
volume-show

**jd volume show --name home/external-efs1 (efs)**
volume-show-efs

**jd volume status**
volume-status

**jd volume status --name home/external-efs1**
volume-status-efs

**jd volume backup (host not stopped)**
volume-backup-running

**jd volume backup (efs - invalid)**
volume-backup-efs

**jd volume backup (host stopped)**
volume-backup

**jd volume backup --all**
volume-backup-all

**jd config --restore-volumes (with az-swap)**
config-az-swap

**subsequent jd up - jd open**
up-open


JGuinegagne:
Full E2E test [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35348166273)

JGuinegagne:
Update E2E [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35362512394) based on pytest marker selector

JGuinegagne:
Another E2E [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35367969405); last run hit timeout issue on SSM execution

michaelnchin (commented):
Thanks @JGuinegagne! Per grill-pr: This looks solid overall. Two comments worth addressing before merge, both on the `ebs_snapshot_ids` map (the `jd down`/`jd up` note in `variables.tf`, and validating map keys at plan time); the rest are non-blocking.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
```diff
@@ -484,4 +490,37 @@ variable "additional_efs_mounts" {
...
+    a key that a live volume was created from plans a replacement of that volume -- recreating it EMPTY
+    and discarding everything on it. The same applies to pointing a key at a different snapshot: the
+    volume is replaced with the contents of the new one. Leaving a stale map in place costs nothing;
+    "tidying" it costs the data. Two consequences worth knowing:
+
+      - after a restore, the map keeps naming snapshots that `jd volume backup` may later supersede and
+        delete. That is harmless while the volumes are not being replaced, but a later replacement fails
+        with a not-found rather than silently emptying the volume -- loud, and recoverable by running
+        `jd config --restore-volumes` again to repoint the map at the current backups;
+      - a `jd down` followed by `jd up` in the same project directory recreates the volumes FROM THIS
```
michaelnchin: `[Reliability]` This bullet says a `jd down` followed by `jd up` recreates the volumes from this map, but `null_resource.reap_volume_backups` (modules/volumes/main.tf:230) deletes every snapshot carrying the DeploymentId on destroy — and `jd down` is `terraform destroy`. So after `jd down` the map names snapshots that no longer exist, and the next `jd up` fails on `aws_ebs_volume` with `InvalidSnapshot.NotFound` (the same failure `test_volume_swaps.py:97-105` describes for a superseded snapshot). The bullet two lines up then tells the user never to clear the map, so the documented recovery path is closed too.

I think the reaper is the right call (a torn-down deployment has nothing to restore into), so the fix is the text: after `jd down` the backups are gone, and `jd up` in the same directory needs `ebs_snapshot_ids` cleared — or `jd down` could clear it. Which do you want?

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/volumes/main.tf
```diff
@@ -1,16 +1,83 @@
...
+
 # Define EBS volume for the notebook data (will mount on /home/jovyan)
 resource "aws_ebs_volume" "jupyter_data" {
   availability_zone = var.availability_zone
   size              = var.volume_size_gb
   type              = var.volume_type
   encrypted         = true
+  # Absent key -> null -> an empty volume, which is the fresh-deployment default. try() rather than
+  # lookup() because lookup's default would have to be "" and aws_ebs_volume wants null.
+  snapshot_id = try(var.ebs_snapshot_ids[local.home_volume_name], null)
```
michaelnchin, line 45: `[Reliability]` An unknown key in `var.ebs_snapshot_ids` is silently ignored: `try(..., null)` yields an empty volume, which is the outcome this whole PR exists to prevent. The CLI path only writes identities from the inventory, but a hand-edited map, a renamed `mount_point` (the identity is `home/<mount_point>`, so renaming a mount orphans its key), or a drift between the manifest's `name: home` (manifest.yaml:60) and `local.home_volume_name` here all produce a restore that "succeeds" with an empty volume.

Since this resource already carries the cross-variable precondition, a second one would close it at plan time:
```hcl
precondition {
  condition = alltrue([
    for key in keys(var.ebs_snapshot_ids) :
    contains(concat([local.home_volume_name], values(local.additional_volume_names)), key)
  ])
  error_message = "ebs_snapshot_ids names a volume this configuration does not create: ${join(", ", setsubtract(keys(var.ebs_snapshot_ids), concat([local.home_volume_name], values(local.additional_volume_names))))}. Keys must be 'home' or 'home/<mount_point>' of an additional_ebs_mounts entry."
}
```

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/volumes/main.tf
```diff
@@ -146,4 +214,66 @@ resource "aws_efs_mount_target" "additional_efs_targets" {
...
+  provisioner "local-exec" {
+    when        = destroy
+    interpreter = ["/bin/bash", "-c"]
+    command     = <<-DOC
+      set -uo pipefail
+
+      SNAP_IDS=$(aws ec2 describe-snapshots \
+        --owner-ids self \
+        --region "${self.triggers.region}" \
+        --filters "Name=tag:DeploymentId,Values=${self.triggers.deployment_id}" \
```
michaelnchin, line 266: `[Hardening]` non-blocking: this reaps every self-owned snapshot tagged `DeploymentId=<postfix>`, not just the ones `jd volume backup` made. A user who snapshots the data volume from the console with "copy tags from volume" gets `DeploymentId` copied onto their manual backup, and `jd down` deletes it. `volume.backup` already stamps `Source=jupyter-deploy` (manifest.yaml:509-511), so adding `"Name=tag:Source,Values=jupyter-deploy"` to this filter scopes the reaper to what the CLI created at no cost.

libs/jupyter-deploy/jupyter_deploy/handlers/project/config_handler.py
```diff
@@ -267,6 +268,51 @@ def restore_secrets(
...
+        Raises:
+            BackupsNotReadyError: If the backups cannot be shown to hold the volumes' current contents,
+                or if any managed volume has no backup. Restoring only some volumes would recreate the
+                rest empty, which is the data loss this path exists to prevent.
+            IncompatibleHostStateError: If the host is not stopped, so quiescence cannot be established.
+            KeyError: If the template declares no variable by that name, raised by the record write
+                before the project is touched.
+        """
+        volume_handler = VolumeHandler(display_manager=self.display_manager)
+        volume_handler.validate_backups_ready()
```
michaelnchin, line 301: `[Reliability]` The quiescence proof (host stopped, every backup postdates the last stop) is established here, at `jd config` time, but the volumes are replaced at `jd up`. Anything between the two — `jd host start`, `jd open`, a scheduled job — reopens the window the check closed, and the apply then restores a stale backup with no symptom. The docstring covers a user re-running the check by hand, but not this gap, which is the normal `jd config && jd up` flow with a pause in it.

Is re-validating at `jd up` (when the pending plan replaces a volume with `snapshot_id` set) feasible, or is the `UpHandler` too far from the volume machinery? If not, a closing `hint` from this command — "keep the host stopped until `jd up` completes; starting it invalidates these backups" — would at least name the constraint.

.github/workflows/e2e-reap-stale.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      # Needed by find-takedown-jupyterlab: it restores each project from the S3 store, and the
+      # store credentials live in the CI project.
+      - name: Restore CI project
+        run: just ci-restore sandbox-ci
+
+      - name: Reap stale deployments
+        run: |
+          just reap-stale-jupyterlab \
+            "${{ inputs.older-than-hours || '12' }}" \
```
michaelnchin: `[Hardening]` non-blocking: `${{ inputs.older-than-hours }}` and `${{ inputs.dry-run }}` are interpolated into a `run:` script. Only write-access actors can dispatch, so the blast radius is small, but the usual pattern avoids it entirely:
```yaml
env:
  OLDER_THAN_HOURS: ${{ inputs.older-than-hours || '12' }}
  DRY_RUN: ${{ inputs.dry-run && '--dry-run' || '' }}
run: just reap-stale-jupyterlab "$OLDER_THAN_HOURS" sandbox-e2e $DRY_RUN
```

scripts/reap_stale_jupyterlab.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def find_stale_deployments(older_than_hours: float, region: str | None = None) -> list[tuple[str, float]]:
+    """Return (deployment_id, age_hours) for every jupyterlab instance older than the threshold."""
+    cmd = [
+        "aws",
+        "ec2",
+        "describe-instances",
+        "--filters",
+        f"Name=tag:Template,Values={TEMPLATE_TAG}",
```
michaelnchin, line 100: `[Reliability]` non-blocking question: the three gates (credentials, `Template` tag, age) mean any `tf-aws-ec2-jupyterlab` instance in the E2E account older than 12h is destroyed — including a future canary slot, a release-gate run someone is inspecting, or a maintainer's manual deployment in that account. Worth an explicit opt-out (skip instances tagged e.g. `KeepAlive=true`), so "don't reap this one" is a tag rather than a race against the cron?

AGENT.md
```diff
@@ -160,6 +160,8 @@ Always run from the root of the repository:
    - Runs `ruff format`, `ruff check --fix`, `mypy`, `terraform fmt`, and `yamllint`
 2. Run unit tests: `just unit-test`
    - Runs `uv run pytest`
+3. If you added a CLI command or edited any CLI docstring, regenerate the reference docs: `just docs-cli-ref`
```
michaelnchin, line 163: `[Review-risk]` non-blocking: the "EC2 JupyterLab template package" section doesn't record what this PR adds to that package — the `volumes:` manifest contract, that `jd volume backup` creates snapshots outside terraform state and `jd down` reaps them by `DeploymentId`, and that the operator now needs `ec2:CreateSnapshot`/`DescribeSnapshots`/`DeleteSnapshot` and `efs:DescribeFileSystems` (CI runs as `AdministratorAccess`, so the e2e suite can't catch a missing grant). The template README's "AWS credentials are the only prerequisite" is still true, but "any credentials" no longer is.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
JGuinegagne: Good callout. After a successful `jd down` there should be no billable resource left, so no EBS snapshot. I'll adjust the comment.

libs/jupyter-deploy/jupyter_deploy/handlers/project/config_handler.py
JGuinegagne: That's a good point. In spirit, `jd up` shouldn't re-plan, otherwise we are conflating `jd config` & `jd up`. Similarly, `jd config` should be "read-only", not have mutation via side-effect; that's why it doesn't create snapshots, just finds them. I'm a bit wary of breaking that mental model either way. Adding an hint at the end of `jd config --restore-volumes` is a good proposal though!

AGENT.md
JGuinegagne: Ah good point, this [README](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-ec2-jupyterlab/README.md) needs an inventory of resources, like the [EKS OIDC](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-eks-oidc/README.md) or [base](https://github.com/jupyter-infra/jupyter-deploy/blob/main/libs/jupyter-deploy-tf-aws-ec2-base/README.md) template README.

We'll pick this up as part of the documentation task of the `jupyterlab` template (cc @LahariChowtoori ), and we should also add explanation at that point in the `AGENT.md.template`.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/volume_handler.py
```diff
@@ -0,0 +1,485 @@
...
+
+        Normally zero or one candidate; sorting covers the transient window inside backup_volume(),
+        between creating the replacement and deleting the one it supersedes.
+        """
+        if not volume_name:
+            return None
+        matching = [b for b in backups if b.get("volume_name") == volume_name]
+        if not matching:
+            return None
+        return sorted(matching, key=lambda b: str(b.get("created_at", "")), reverse=True)[0]
```
bhanderig, line 282: `[Reliability]` The restore map can resolve to a snapshot that can never seed a volume.

`state` is in the payload — `_describe_snapshots` emits it (`aws_ec2_runner.py`) and `show_volume` renders it as `backup_state` — but neither this sort nor `resolve_backup_ids` reads it.

Concretely: `jd volume backup` creates S2, `wait_snapshot_completed` sees `error` and raises, so the good older S1 is correctly *not* deleted. S2 still has the newer `StartTime`, so it wins here. `jd config --restore-volumes` writes the errored id into `ebs_snapshot_ids`, and `validate_backups_ready` passes it because it only checks the timestamp against the last stop. `jd up` then plans a replacement — `aws_ebs_volume` has no `create_before_destroy` — so the live volume is destroyed before `CreateVolume` fails on an unusable snapshot, while S1 sat right there.

Same shape via the 900s default in `wait_snapshot_completed`: a volume large enough to exceed it leaves `pending` as the newest.

Suggest filtering to `state == completed` here (and/or in `resolve_backup_ids`), so an errored or in-flight snapshot is never resolvable and the last known-good backup wins.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/volume_handler.py
```diff
@@ -0,0 +1,485 @@
...
+        if not volume.backups_map:
+            kind = f"'{volume.volume_class}'" if volume.volume_class else "this class of"
+            raise VolumeNotBackupableError(
+                name,
+                f"the template declares no backup mechanism for {kind} storage",
+                volume_class=volume.volume_class,
+                hint="Run 'jd volume backup --all' to back up every volume that supports it.",
+            )
+
+        superseded = [
```
bhanderig, line 386: `[Reliability]` The supersede set is computed from volume identity alone and never consults the backups map — so `jd volume backup` can delete the snapshot `ebs_snapshot_ids` is currently pointing at.

`test_volume_swaps.py:85-110` already documents this biting in practice:

> "On a re-run that map is left over from the previous swap and may name a snapshot that `jd volume backup` has since superseded and DELETED — `test_volume.py` alone does that twice to the home volume. The apply then dies with `InvalidSnapshot.NotFound` halfway through, having already destroyed the volume it could not recreate."

The fixture's answer is to re-resolve before applying. Nothing does that for a user, and the four steps are all things a careful one does: backup → `--restore-volumes` → remember one more thing, start/stop/backup again → `jd up`, which now destroys the live volume and fails to recreate it.

The handler can already read the map — `_resolve_value` handles `ValueSource.TEMPLATE_VARIABLE`, and `manifest.yaml` wires `volume_backup_ids` → `ebs_snapshot_ids` — so skipping any `superseded_id` the map still names is a couple of lines and removes the hazard rather than documenting it.

Worth noting the recovery `variables.tf` points at may also be unavailable: `resolve_backup_ids` → `_inventory()` → `_resolve_value("jupyter_data_volume_id")` reads an output whose resource was just destroyed, so if terraform doesn't retain the stale value the user gets `Volume information is not available… :bulb: Deploy the project with 'jd up' first` — a refusal naming the command that just failed. Distinct from the config→up window in your `config_handler.py:301` thread: that one restores a stale backup silently, this one fails loudly with the volume already gone.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/volumes/main.tf
```diff
@@ -146,4 +214,66 @@ resource "aws_efs_mount_target" "additional_efs_targets" {
...
+        --query 'Snapshots[].SnapshotId' \
+        --output text 2>&1) || {
+        echo "ERROR: could not list volume backups for deployment ${self.triggers.deployment_id}: $SNAP_IDS" >&2
```
bhanderig: `[Reliability]` This makes `jd down` fail for deployments that never took a backup.

The describe runs unconditionally on every teardown — the "no backups to reap" fast path is below it — so `ec2:DescribeSnapshots` becomes a hard requirement for destroying every existing deployment, including ones where `jd volume backup` was never invoked. The operator gets a failed destroy and a half-torn-down stack, and the error text sends them to `ec2:DeleteSnapshot`, which isn't the permission that failed.

The comment's reasoning — "a delete that fails must fail the destroy… what makes failing safe is idempotency" — holds for `delete-snapshot`. It doesn't transfer here, where the failure isn't a leaked snapshot but a blocked teardown on a deployment with nothing to leak.

Separate from michaelnchin's `AGENT.md:163` thread on documenting the new grants: even once they're documented, this ordering means the *describe* is what breaks teardown. Suggest distinguishing authorization failures from the rest, or gating the strictness on `length(var.ebs_snapshot_ids)` as a trigger. At minimum the message should name `ec2:DescribeSnapshots`.

scripts/reap_stale_jupyterlab.py
```diff
@@ -0,0 +1,143 @@
...
+            # An instance carrying the Template tag but no DeploymentId cannot be reaped safely:
+            # the takedown path is keyed on that id, so there is nothing to hand it.
+            print(f"  ! skipping an instance with no DeploymentId tag (launched {launch_time})")
+            continue
+
+        age_hours = (now - datetime.fromisoformat(launch_time)).total_seconds() / 3600
+        if age_hours > older_than_hours:
+            # Keep the OLDEST age seen per deployment: a deployment whose instance was replaced
+            # mid-run has a young instance, and reporting the young one would understate the leak.
+            stale[deployment_id] = max(stale.get(deployment_id, 0.0), age_hours)
```
bhanderig: `[Reliability]` The reaper is weakest against exactly the runs it was added to catch, and this `max()` guards a case that can't arise.

Age comes from `LaunchTime`, which EC2 updates on start — and `test_volume_swaps.py:452` explicitly asserts the zone swap *replaced the instance* (`assert cli.get_str_output("instance_id") != instance_id_before`). So a run that fails at or after the swap leaves a brand-new instance and the 12h clock restarts from that moment. Keep-on-failure plus a late-stage failure is the scenario this file exists for, and it's the one it can't see.

On the `max()` itself: an AZ swap destroys the old instance, so there's only ever one age to compare. And `if age_hours > older_than_hours` above has already filtered the young instance out, so the "instance replaced mid-run" case the comment describes can't reach this line either way.

Creation time from the S3 store entry, or a tag stamped at apply, would survive an instance replacement in a way `LaunchTime` doesn't.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/configurations/env.example
```diff
@@ -31,6 +31,12 @@ E2E_DOCKERFILE=
 # per account (us-west-2c in one, us-west-2d in another) — and a GPU type offered in one is not
 # necessarily offered in the other: g5.xlarge does not exist in us-west-2d at all.
 JD_E2E_AVAILABILITY_ZONE=us-west-2a
+# The zone `test_volume_swaps.py` moves the deployment INTO, to prove a zone change preserves every
+# volume's data. Must differ from JD_E2E_AVAILABILITY_ZONE above, must have a default-VPC subnet
+# (the template selects its subnet by zone), and must offer JD_E2E_CPU_INSTANCE — the swap replaces
+# the instance, so a type with no capacity there fails the apply rather than the assertion.
+# Check with: aws ec2 describe-subnets --filters Name=default-for-az,Values=true
+JD_E2E_ALT_AVAILABILITY_ZONE=us-west-2b
```
bhanderig, line 39: `[Review-risk]` This default can't be verified from the PR and fails at the release gate.

CI takes it verbatim — `e2e-jupyterlab-job.yml:107` is a bare `just env-setup-jupyterlab` with no overrides. Two things make that risky, both already on the record here:

1. The comment 8 lines above says zone names aren't portable across accounts — "the same zone NAME maps to different hardware per account." So `us-west-2b` working in whichever account produced the 123/0 run isn't evidence about the CI account.
2. This workflow doesn't run on PR checks. Per the `g5.xlarge`-in-`us-west-2d` thread on #373, it surfaces only when `release-jupyterlab.yml` runs it as the gate for `prod-pypi-publish` — after the fact.

If the CI account's default VPC has no `us-west-2b` subnet, or `t3.medium` isn't offered there, the swap fails at `jd up` rather than at an assertion — and under this PR's new teardown rule a failed run keeps the deployment, so the first symptom is a leaked deployment plus a red release gate.

Could you post the output of these two, with CI credentials?

```
aws ec2 describe-subnets --filters Name=default-for-az,Values=true \
  --query 'Subnets[].AvailabilityZone' --region us-west-2
aws ec2 describe-instance-type-offerings --location-type availability-zone \
  --filters Name=location,Values=us-west-2b Name=instance-type,Values=t3.medium --region us-west-2
```

Alternatively, have the workflow pass `alt-availability-zone=` explicitly, which is where #373 landed for the GPU type.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/volume_handler.py
```diff
@@ -0,0 +1,485 @@
...
+
+        An unknown stop time is a refusal, not a pass. EC2 does not always report one, and "cannot
+        prove" must not read as "proved safe" for an operation that destroys the volumes.
+
+        Raises:
+            BackupsNotReadyError: If quiescence cannot be established for every volume, or if any
+                managed volume has no backup at all.
+            IncompatibleHostStateError: If the host is not stopped.
+        """
+        if not self.project_manifest.supports_volume_readiness():
```
bhanderig, line 454: `non-blocking:` three smaller items, grouped to keep the thread count down.

**1. `[Hardening]` The readiness check is a silent no-op when the template omits the command (this line).** `manifest.py` states the intent plainly — "a template that omits it is checked for nothing" — and that's a defensible opt-out. The gap is that nothing validates the one combination where it's dangerous: a `volumes:` entry declaring a `backups-map` (so `--restore-volumes` will replace volumes) while `volume.validate-backups-ready` is absent (so quiescence is never established). `manifest_validation.py` is where that belongs and already gained 45 lines of cross-cutting checks here. Check (1) survives either way since `config_handler.restore_volumes` calls `resolve_backup_ids()` itself — it's (2) and (3) that vanish without a word.

**2. `[Hardening]` `verify_instance_stopped_since` passes vacuously on an empty list (`ec2_instance.py:225`).** `raw.split(",")` on `""` yields `[""]`, which parses to nothing, so `since == []`, so `stale == []`, so the gate returns clean. The runner docstring already owns the reasoning — "dropping one here can only make this check pass more easily" — and `validate_backups_ready`'s `undated` branch guards today's only caller. Still, a gate that can't tell "nothing to check" from "verified" is one `raise` away from being able to.

**3. `[Architecture]` `VOLUME_STATE_COMMAND = "volume.list"` collides with `jd volume list`, which never calls it (`manifest.py:44`).** `list_volumes()` makes no cloud call, as its own docstring says; the manifest command named `volume.list` is the live-state lookup, consumed by `show_volume()` and `get_status()`. So the one command whose name matches a CLI verb is the one that verb doesn't use, and `has_command("volume.list")` reads as "does this template support listing volumes" when it means "can it report zone/capacity/state." `volume.describe` or `volume.state` costs one rename across `manifest.py` + `manifest.yaml`.

scripts/reap_stale_jupyterlab.py
JGuinegagne: Good call, a `KeepAlive` opt-out is the right shape. Deferring until we set up canaries, since that is when a long-lived deployment in this account first needs protecting.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/volumes/main.tf
JGuinegagne: Getting as `length(var.ebs_snapshot_ids)` makes sense.

Generally, we should display the error message rather than swallow it, so that we don't have to distinguish permission errors.

scripts/reap_stale_jupyterlab.py
JGuinegagne: sure. It's a fairly minor point, but worth hardening.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/tests/e2e/configurations/env.example
JGuinegagne: right, but 1/ I tested that w/ runs on the CI accounts, 2/ default VPCs def have subnets in `us-west-2b`, 3/ `t3.medium` definitely _is_ offered in `us-west-2` (our CI has been running for months!).

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
```diff
@@ -48,8 +48,14 @@ variable "availability_zone" {
...
     offered in every zone, and even where offered it can be refused for lack of capacity.
 
-    Changing this on an existing deployment is NOT supported: EBS volumes cannot cross zones, so
-    the data volume would have to be snapshotted and restored. Choose it at deployment time.
+    Changing this on an existing deployment DESTROYS the data on every EBS volume unless you back
+    them up first: a block volume cannot cross zones, so terraform must replace it. Run
+    `jd volume backup --all`, then `jd config --restore-volumes --availability-zone <zone>`, and
+    each volume is recreated from its backup.
+
+    NOTHING STOPS YOU FROM SKIPPING THAT. There is no plan-time check: the only guard is inside
```
andrii-i, line 56: We discussed a `jd config` guard for this in https://github.com/jupyter-infra/jupyter-deploy/pull/373#discussion_r4026432597. Should we track it in an issue so it wouldn't get lost?

.github/workflows/e2e-jupyterlab-fresh.yml
```diff
@@ -207,20 +215,56 @@ jobs:
...
+  # runs, and do not loosen the reaper without reverting this.
   destroy:
     name: "Teardown deployment"
-    needs: [build-image, deploy, test, test-mutating]
-    if: always() && inputs.destroy-after && needs.deploy.outputs.deployment-id != ''
+    needs: [build-image, deploy, test, test-gpu, test-mutating]
+    if: >-
+      always() && inputs.destroy-after
+      && needs.deploy.outputs.deployment-id != ''
+      && needs.deploy.result == 'success'
```
andrii-i, line 264: If the deploy job fails after creating resources, they are left running and no re-run can reuse them ([example run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35362512394)). Should we tear them down when deploy fails, as before, and keep the deployment only when a test fails?

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
```diff
@@ -0,0 +1,118 @@
...
+        VolumeId=volume_id,
+        Description=description or f"jupyter-deploy backup of {volume_id}",
+        TagSpecifications=[{"ResourceType": "snapshot", "Tags": tag_specs}],
+    )
+
+
+def wait_snapshot_completed(
+    ec2_client: EC2Client,
+    snapshot_id: str,
+    timeout_seconds: int = 900,
```
andrii-i, line 55: Backing up a big disk can take longer than the 15 minutes this waits. When time runs out the command crashes with debug prints and no clear human-readable error; nothing is lost, the backup keeps going on its own, but the user is not told that or what to do next. We should at least print a clear error with the failure reason, the backup's name, how to check on it, and the next step, and raise the limit well above a first snapshot.

The larger fix is to stop blocking on completion. Three ways tools handle this: 1/ return as soon as the snapshot exists and let the user poll, with an optional `--wait` (`aws ec2 create-snapshot`, `velero backup create`); 2/ keep blocking but with no fixed limit, a progress display, and a re-run that re-attaches to the pending snapshot instead of creating another (`kubectl rollout status` style); 3/ hand backups to AWS Backup or Data Lifecycle Manager and only resolve the latest completed one at restore. I would go with 1/: a snapshot's content is fixed the moment it is requested, so waiting buys nothing for the data, only the ability to delete the old snapshot right away. With restore limited to completed snapshots (the other thread), the old one can be deleted at the next backup or restore instead, `jd volume show` already reports the backup's state, and `--wait` keeps the e2e and scripts unchanged. 2/ still ties up the terminal for hours and needs the same re-attach logic. 3/ is a poor fit: scheduled snapshots run whether or not the host is stopped, so they can miss data still in memory, which is exactly what the stopped-host rule here prevents, and they add another AWS service and its permissions for what is a single deliberate backup before a move.

libs/jupyter-deploy/jupyter_deploy/handlers/project/config_handler.py
```diff
@@ -267,6 +268,60 @@ def restore_secrets(
...
+        creating the backups is the separate, explicit `jd volume backup`.
+
+        Validates before writing anything. Pointing the variable at a backup REPLACES the volume on the
+        next apply -- `snapshot_id` forces replacement -- so this path is destructive whether or not
+        anything else about the configuration changed, and it succeeds silently when the backup is old:
+        terraform recreates the volume from it, reports success, and the work done since is gone. The
+        check is only meaningful here, before the variable is written.
+
+        Scoped to this flag on purpose. A user who changes a zone WITHOUT asking for a restore gets no
+        check from here -- the engine's plan-time precondition is what stands between them and an empty
```
andrii-i: nit: remove the mention of the safety check, as it was removed?

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_ec2_runner.py
```diff
@@ -64,6 +129,53 @@ def _describe_instance_status(
...
+        against a shutdown and knows nothing about what they mean. Unparseable entries are dropped rather
+        than guessed at; a caller that cannot supply a usable instant must refuse on its own, because
+        dropping one here can only make this check pass more easily.
+        """
+        instance_id = require_arg(resolved_arguments, "instance_id", StrResolvedInstructionArgument).value
+        raw = require_arg(resolved_arguments, "timestamps", StrResolvedInstructionArgument).value
+
+        since = [parsed for parsed in (str_utils.parse_timestamp(entry.strip()) for entry in raw.split(",")) if parsed]
+
+        self.display_manager.info("Verifying the backups predate the last shutdown")
```
andrii-i: nit: postdate?

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/modules/volumes/variables.tf
```diff
@@ -57,4 +57,13 @@ variable "instance_id" {
...
   description = "ID of the security group for EFS mount targets."
   type        = string
-}
\ No newline at end of file
+}
+variable "ebs_snapshot_ids" {
+  description = <<-EOT
+    Map of volume name -> EBS snapshot id to create that volume from.
+
+    Key "home" is the jupyter data volume; other keys match the "name" of an additional_ebs_mounts
```
andrii-i: nit: `home/<mount_point>`?

andrii-i (commented):
Thanks @JGuinegagne. Looks good overall. Before this goes in it would be good to resolve the two problems that can lead to data loss: [a restore picking an unfinished or failed backup over a good one](https://github.com/jupyter-infra/jupyter-deploy/pull/381#discussion_r4051098452) and [a new backup deleting the snapshot a pending restore still points at](https://github.com/jupyter-infra/jupyter-deploy/pull/381#discussion_r4051098599). Both look like quick fixes. The rest are nits, fine for a follow-up.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
andrii-i: Reproduced this. Since `jd down` deletes the backups, should `jd down` say so explicitly before its confirmation prompt?

libs/jupyter-deploy/jupyter_deploy/handlers/resource/volume_handler.py
andrii-i: +1, reproduced this as well.

libs/jupyter-deploy/jupyter_deploy/handlers/resource/volume_handler.py
andrii-i: +1, seeing this as well.

libs/jupyter-deploy/jupyter_deploy/handlers/project/config_handler.py
andrii-i: `jd config --restore-volumes` could record, in the plan metadata it already writes for `jd up`'s progress counts, that this plan restores volumes from backups; `jd up` could then re-run the stopped-since-backup check before applying. A read-only check, no second plan. That would enforce the "keep the host stopped until `jd up` completes" rule instead of only telling the user about it in the hint.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
JGuinegagne: yeah, I went back and forth a few times during implementation on whether to add a guard to `jd config`, and landed on not changing the core `jd config` for now.

We definitely _should_ add a warning to prevent data loss, but there are other destructive cases (e.g. remove additional volumes, change aws region, etc) that we should cover as well. I'll track as an issue.

.github/workflows/e2e-jupyterlab-fresh.yml
JGuinegagne: hum, good point if we can't retry the deploy step, but i'm not sure that's the case. And even if that's the case, it would be better to update the plugin to make the deploy step retryable. Would have to be a follow up issue as well.

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
JGuinegagne: Thanks Andrii. Is EBS performance that bad that it would take > 15 mins to restore say a 1TB backup? I don't have a strong intuition w.r.t. performance throughput, but it would have to be lower than 1GB/s to exceed the threshold. That said, we are talking about upload speed here, so it may genuinely be below that threshold. Also note: rn the default EBS volume sizes are 30GB, so well below problematic levels.

But yeah, at a high level a `--wait` could make sense here. I'm just worried about introducing this pattern in the CLI just yet. How about we just raise a  new `ResourcePollTimeoutError` instead that tells customer to check `jd volume show` for completion?

libs/jupyter-deploy/jupyter_deploy/provider/aws/aws_ec2_runner.py
JGuinegagne: haha thanks, nice catch!

JGuinegagne:
E2E tests [run](https://github.com/jupyter-infra/jupyter-deploy/actions/runs/35415506869) with all comment-related changes

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
```diff
@@ -0,0 +1,127 @@
...
+                "describe_snapshots returned no snapshot while waiting for it to complete",
+            )
+
+        snapshot = snapshots[0]
+        state = snapshot.get("State", "")
+
+        if state == SNAPSHOT_COMPLETED_STATE:
+            return snapshot
+        if state == SNAPSHOT_ERROR_STATE:
+            raise RuntimeError(f"Snapshot {snapshot_id} entered the 'error' state and cannot be used.")
```
andrii-i, line 85: still a bare `RuntimeError`, so a snapshot that AWS marks `error` ends in a Python traceback instead of a message, like the timeout did before `ResourcePollTimeoutError`, nonblocking

andrii-i (approved):
Thanks @JGuinegagne for addressing the feedback and especially the data-loss-related problems. There are a couple of unresolved threads but they are non-blocking.

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
andrii-i: Depends on how much data the volume holds. AWS publishes no upload rate for snapshot creation; the only documented rate is on the restore side, [100 to 300 MiB/s](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html). At that rate the template's default 30 GB volume takes 2 to 5 minutes, a 500 GiB one 30 to 85 minutes, and `volume_size_gb` has no upper bound, so both cases are real.

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
JGuinegagne: Thanks! This one is acceptable as a stack trace I think (something genuinely went wrong). Alternatively we need a new dedicated error, I'll sleep on it.

libs/jupyter-deploy-tf-aws-ec2-jupyterlab/jupyter_deploy_tf_aws_ec2_jupyterlab/template/engine/variables.tf
JGuinegagne: #386

.github/workflows/e2e-jupyterlab-fresh.yml
JGuinegagne: #387

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
JGuinegagne: #388

libs/jupyter-deploy/jupyter_deploy/api/aws/ec2/ebs_snapshot.py
JGuinegagne: #388

